PR #64, opened 2015-12-29
Issue #59 Invalid (non-aasembly) file was shown as assembly.

iCodeIT:
This is a tempory fix. When the code is fixed in the nUnit engine this
code shall be removed.


src/nunit-gui/Presenters/TestPropertiesPresenter.cs
```diff
@@ -118,6 +114,19 @@ private void DisplayTestProperties()
                 _view.Header = testItem.Name;
             }
         }
+
+        public static string GetTestType(TestNode testNode)
+        {
+            if (testNode.RunState == RunState.NotRunnable)
```
CharliePoole: This also needs to be conditioned on testNode.Type == "Assembly" A non-runnable Fixture, for example, is still a Fixture.

src/tests/Presenters/TestPropertiesPresenterTests.cs
```diff
@@ -0,0 +1,63 @@
...
+    [TestCase("test.exe", "Skipped", ExpectedResult = "Assembly")]
+    [TestCase("test.exe", "Unknown", ExpectedResult = "Assembly")]
+    [TestCase("test.png", "Runnable", ExpectedResult = "Assembly")]
+    [TestCase("test.png", "NotRunnable", ExpectedResult = "Unknown")]
+    [TestCase("test.png", "Ignored", ExpectedResult = "Assembly")]
+    [TestCase("test.png", "Explicit", ExpectedResult = "Assembly")]
+    [TestCase("test.png", "Skipped", ExpectedResult = "Assembly")]
+    [TestCase("test.png", "Unknown", ExpectedResult = "Assembly")]
+    public string GetTestTypeTest(string fullname, string runstate)
+    {
```
CharliePoole: Good that we have tests! This was all done as a spike, which is why we don't have any up till now. We should add others at some point.

CharliePoole:
Good work. I think it's ready to merge with the change I pointed out. You should also add test cases for one or more types != Assembly. Just push your changes and this PR will update itself.

@nunit/core-team We need a second review here, please.

CharliePoole:
@iCodeIT I'll try to find some other relatively simple ones for you or break down one of the bigger ones into pieces.

CharliePoole:
Looks good now.

CharliePoole:
This fixes #59

@iCodeIT Always have a link back to the issue in the PR description or a comment. That way we can follow the link to the issue or the backlink on the issue to the PR. If you use the form fixes #xx or fix #xx, then the issue is automatically closed when we merge.

PR #70, opened 2016-01-08
equalize different project framework versions to .NET framework 4.0

jaxx:


CharliePoole:
What are you fixing? There is no issue about this and therefore no discussion. If we do this .NET 4.0 will be a prerequisite for running the gui.

CharliePoole:
Also, the engine package only exists for 2.0.

jaxx:
Hi,
when I load the tests I get error "Could not load file or assembly 'nunit.framework' or one of its dependencies. This assembly is built by a runtime newer than the currently loaded runtime and cannot be loaded." from catch part. 
```
try
{
    Runner = _testEngine.GetRunner(package);
    // TODO: Error here when multiple files are run
    Tests = new TestNode(Runner.Explore(TestFilter.Empty));

     _resultIndex.Clear();
}
catch (Exception ex)
{
    //if (TestException != null)
    //    TestException(new TestEventArgs(TestAction.TestLoadFailed, ex));
}
```
My test projects are compiled in .NET framework 4.0. When I upgrade the versions then there's no problem.

jaxx:
It may also help if only supported runtime property is modified in App.config file.

CharliePoole:
Current codebase contains a line added for debug purposes, which forces the test to run in the same process as the GUI itself. See https://github.com/nunit/nunit-gui/blob/master/src/nunit-gui/Presenters/MainPresenter.cs#L280

I hope you are not using this unreleased code for production work!

PR #71, opened 2016-01-08
Fix double parsing

jaxx:
After test run, parsing test duration fails with "Input string was not in a correct format."


CharliePoole:
@jaxx Looks like you may have found a bug and added it as a second commit. Best way to deal with that is to file an issue although we'll take a PR if the description has all the info that would normally be in a bug. However, our review process and change log creation all work better if we have issues.

Additionally, PRs should deal with one thing and not get unrelated commits added.

My guess is that you already figured all of the above out yourself, but I thought I would mention it.

PR #73, opened 2016-01-09
Fixes #69

### LINKED ISSUE #69: Reloading nunit project file causes engine error
This happens because the engine modifies the test package by expanding the project. When reloading, it gives an error if the project is already expanded.

We can fix this in the gui by creating an entirely new package on reload. This is probably best, since the nunit project file might have been modified.

However, this issue points out that problems can arise if the engine modifies structures created by the test runner. This should be looked at in the engine.

iCodeIT:
Moved creation of package into the model in order to allow reload to create new package. The list of files is stored in the model. GuiOptions are passed as parameters, since the options might change between reloads.

### DIFF
--- a/src/nunit-gui/Model/ITestModel.cs
+++ b/src/nunit-gui/Model/ITestModel.cs
@@ -107,13 +107,13 @@ namespace NUnit.Gui.Model
         void SaveProject();
 
         // Load a TestPackage
-        void LoadTests(TestPackage package);
+        void LoadTests(IList<string> files, GuiOptions options);
 
         // Unload current TestPackage
         void UnloadTests();
 
         // Reload current TestPackage
-        void ReloadTests();
+        void ReloadTests(GuiOptions options);
 
         // Run just the specified ITestItem
         void RunTests(ITestItem testItem);
--- a/src/nunit-gui/Model/TestModel.cs
+++ b/src/nunit-gui/Model/TestModel.cs
@@ -98,6 +98,8 @@ namespace NUnit.Gui.Model
         }
 
         private ITestItem _selectedTest;
+        private IList<string> _files;
+
         public ITestItem SelectedTest
         {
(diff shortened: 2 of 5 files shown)


CharliePoole:
@iCodeIT I just finished a big refactoring. Don't worry about resolving conflicts for now, I'll pull your code into my branch and see how it fits.

CharliePoole:
I merged it in to my local master and it reloads NUnit projects quite nicely. I  want to think about the pros and cons of passing the GuiOptions into the model, which heretofore didn't know about them.

I'll sleep on it!

iCodeIT:
Actually I think GuiOptions is the wrong name. It contains information on how to run the tests, some fileinformation, which configuration is active and so on. It really has nothing to do with the presentation or view layers and I think it belongs in the model.
GuiOptions would contain stuff like font size, background color, use checkboxes and so on. (It does have a memeber "ShowHelp" that I think is GUI related - maybe that should be in a separate class?).

CharliePoole:
You may not dug deeply enough to see that GuiOptons represents the command-line options for the run. In V2, the console and gui runners were in the same solution and so we used ConsoleOptions and GuiOptions to distinguish them. We can change the name to make that clearer.

In this gui, It is currently processed by the Main and by the MainPresenter. The model is probably a better place to put it, since it can be referenced there by any presenter. However I don't think we want to pass it in with every call but inject it in the  constructor. I'll make a few changes and push it out to a branch where you can review it.

CharliePoole:
@iCodeIT Let's continue the discussion on PR #74.

It incorporates substantial changes both underneath and on top of your changes. Thinking it over, I decided you are right that the command-line options really relate more to the model than the presenter and so put it there.

Please review.

iCodeIT:
Right - I didn't know the history of this. Thanks for "history lesson" - it is always nice to get some background on things.

PR #74, opened 2016-01-14
Issue 69

### LINKED ISSUE #69: Reloading nunit project file causes engine error

CharliePoole:
This PR fixes #69.

* Reapplied @iCodeIT 's changes on top of the updated master.
* Fixed Conflicts
* Renamed GuiOptions to CommandLineOptions
* Gave the model responsibility for handling the command-line options.
* Gave the model responsibility for loading and running tests on startup.
* Simplified the calling sequences for loading and running tests

### DIFF
--- a/src/nunit-gui/Model/ITestModel.cs
+++ b/src/nunit-gui/Model/ITestModel.cs
@@ -73,6 +73,10 @@ namespace NUnit.Gui.Model
 
         #region Properties
 
+        CommandLineOptions Options { get; }
+
+        IRecentFiles RecentFiles { get; }
+
         bool IsPackageLoaded { get; }
 
         // TestNode hierarchy representing the discovered tests
@@ -100,6 +104,9 @@ namespace NUnit.Gui.Model
 
         #region Methods
 
+        // Perform initial actions on startup, loading and possibly running tests
+        void OnStartup();
+
         // Create a new empty project using a default name
         void NewProject();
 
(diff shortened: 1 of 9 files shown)


src/nunit-gui/Model/ITestModel.cs
```diff
@@ -73,6 +73,10 @@ public interface ITestModel : IServiceLocator, System.IDisposable
 
         #region Properties
 
+        CommandLineOptions Options { get; }
```
CharliePoole: Maybe this should be called CommandLineOptions, CommandOptions, RunOptions or something else more specific.

iCodeIT: The options can be both applied from gui and commandline? I would go with something more generic like RunOptions or just Options unless we need more option-classes?

iCodeIT:
I think this looks good! You decide if the name of the options class should be changed. You have a "little" bit more insight ;-)

src/nunit-gui/Model/ITestModel.cs
CharliePoole: No, GuiOptions is purely for the commandline. But its name probably led to your confusion - which is what bad names do. Just Options is more  generic, but then the Type is CommandLineOptions so that's probably OK.

CharliePoole:
In that case, I'll merge it.

PR #78, opened 2016-01-19
Merge StatusBar view into main view and combine presenters

CharliePoole:
**Do Not Merge Yet**

@rprouse @iCodeIT I'd like your comments, primarily on the overall direction I'm taking, exemplified by this change. Design info follows.

The new gui is written as an MVP (Model-View-Presenter) application, adapted to my personal style.

That personal style of gui-creation has always been based on independent controls, each doing their own thing. Usually, I would create a layer of derived controls, where necessary, in order to add basic capabilities to the control, and then top it off with a further layer that understood the domain.

For example, I might have a MultiColorProgressBar, derived from ProgressBar, and then derive a TestProgressBar on top of that.

I find this a nice way to work but it's very difficult to test in an automated way.

I got interested in MVP a few years ago. I used it to write the ProjectEditor and also back-ported some ideas into the V2 gui. And I'm using it for the 3.0 gui, of course.

That gets us to this change...

I originally decided that the main NUnit display form was too complex to have a single view and presenter. So I created multiple, independent views and presenters, as you can see in the setup in Program.cs. Currently (before this change) we have six different views and presenters for the main form. I liked the idea of this separation, but it has several problems:

1. You have to look in several places to understand what happens as a result of an event.

2. Since the presenters don't communicate with one another, any action initiated in one of the sub-views but which needs to be reflected in a different sub-view, has to go through the model. As an example, that's why the model maintains a value for the currently selected test in the tree.

What I'm thinking of doing now is to merge all the views for the main form. Since they already exist separately, I'll just give the main view a property for each sub-view. In the case of the current change, that's the StatusBar property. For the presenter, I will physically merge code from each sub-presenter into the main presenter. As a side benefit, this puts all the updates caused by one event into a single block of code. Note that views and presenters for other windows than the main form are not being changed.

The current commit does this for the StatusBar view and presenter. StatusBarView becomes a child of the Main view and StatusBarPresenter and StatusBarPresenterTests are eliminated, with their code now included in MainPresenter and its tests.

I'd like your review to especially focus on this overall design direction, of course noting any details that strike you as well.

If I go ahead with this, I'll do essentially the same thing for ProgressBar, TestTree, TestProperties and XmlView.


iCodeIT:
From a pure coding point of view I am sure your changes will work.
From a more architectural standpoint I like the small independent building blocks. They are easy to test independently and remove / change / insert very cleanly.
Some of the MVVM frameworks for WPF solves the communication problem between VM's by providing a message service. Basically you can send a message, and anyone interested can receive the message and act on it - it is a kind of advanced observer pattern. Maybe something similar could be used in the MVP pattern for communication between the P's?

Other than the above comments I am sure your proposed change can work as long as the number of GUI elements doesn't grow too large.

CharliePoole:
@iCodeIT It seems to be generally accepted that MVVM works poorly with Windows Forms, which is what led me away from it. MVP is usually described as having two major variants - one in which the views know about the model and one in which everything is run through the presenter. I chose the second because it's the model behind the Presenter-First development approach, which I was using a the time.

I've developed systems like what you describe in the past, usually calling the message service an EventManager. It works well as a global message bus. But it seems like we would be making the app more complex rather than simpler - unless I misread you, that is.

iCodeIT:
I didn't mean to use MVVM - I just refered to it as I have more experience with that, than with MVP. 
Still, I like the smaller building blocks. I agree that a message bus might make a small app like this more complicated instead of less complicated.
Another option is to inject an interface from one presenter into another presenter in order to either subscribe to events from the injected presenter. That way no other presenter needs to know about the statusbar or if the statusbar one day is replaced by a more elaborate status window. Also I like that I can dive right into a component and fix something without having to understand everything else, because everything else i placed elsewhere.
So by now it should be clear that my preference is for small building blocks, but either way will work as long as the gui doesn't get too complex. Your call.

CharliePoole:
Keeping things small and localized is definitely a worthwhile objective and it's what I was after. I was afraid that the complete independence of presenters might make it harder to understand the app, but you don't appear to have had any problem getting into it.

Perhaps I can find another way to keep the separate views and presenters while still allowing a very few messages to be shared.

iCodeIT:
Keeping things independent has made it very easy to dive into the right spot and do stuff. During that process it was easy to understand how that particular component works, and I assume the others have similar patterns. It is true that if you try to understand the whole thing at once it might be more difficult, but starting out making small changes as I did, you actually quite quickly see how things are connected.

CharliePoole:
Actually, this PR was helpful in that it led to the following solution, which I'll implement on another branch...

1. Change the fields in IMainView that refer to the sub-views to use the appropriate interfaces. This willl allow NSubstitute to work with them for testing.

2. Simply let the main presenter handle those view actions that are triggered by one sub-view and reflected in another one!

I'm closing this PR.

PR #79, opened 2016-01-21
Changes to selected test processing

CharliePoole:
The selected test node in the tree did not really belong in the model as it's part of the presentation logic. This PR  takes care of the problem.

* Created more event handler delegates and event argument types, so events will be a bit more type-safe than they were. Discovered some errors in the code in the process.

* Eliminated SelectedTest as a model property.

* Added a model method NotifySelectedTestItemChanged, which causes the model to re-broadcast a change discovered by one of the presenters so that other presenters can handle it.

* Each presenter that uses the selected item independently maintains a field with the current selection.

I plan to follow the rule of three here. That is, if we find a second event that needs to be handled by multiple presenters, I'll just add an event and notify method to the model. But once we find a third one, I would expect to add a subsystem that allows registering events to be broadcast.

### DIFF
--- a/src/nunit-gui/Model/ITestItem.cs
+++ b/src/nunit-gui/Model/ITestItem.cs
@@ -32,7 +32,7 @@ namespace NUnit.Gui.Model
     /// <summary>
     /// ITestItem is the common interface shared by TestNodes
     /// and TestGroups and allows either to be selected in
-    /// the settingsServiceServiceService.
+    /// the tree.
     /// </summary>
     public interface ITestItem
     {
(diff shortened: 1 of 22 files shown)


CharliePoole:
@iCodeIT Would you take a look at this?

iCodeIT:
I think it looks good - one word of warning: I am out with a cold, so I have not been as thorough as I'd like. (there were a lot of code to cover) 
And I like the rule of three.

CharliePoole:
At this stage of the game, we don't have to be terribly good about reviews. I will want to tighten up as we approach a release and follow our normal standards. For now, I'm merging this.

PR #80, opened 2016-01-22
Implement Checked tests execution

### LINKED ISSUE #48: Support checkboxes in the gui tree display
I'm spinning this issue off from #36.

NUnit 2.x uses checkboxes, turned on at the option of the user as a way of selecting tests to be run. It also supports selecting a single node in the tree.

The VS adapter does not use checkboxes. Instead, it offers the standard ctrl and shift approach to selecting multiple items. Additionally, it supports various ways of displaying the tree, so that certain multiple selections are more easily acheived. For example, if tests are grouped by category, it's easy to select the node containing a particular category. If they are grouped by outcome, it's easy to rerun all failed tests.

NUnit 3.0 supports a wide range - even wider than the VS adapter - of ways to organize the display of tests. That being the case, we may not need checkboxes. OTOH, we may want to support them for users who prefer this approach to selection.

If we do support them, we will need to decide if they should be supported in all displays or only in the standard NUnit tree display, which is like that of NUnit V2.

I'm categorizing this as an "Idea" so that we can think about it for a bit after the initial release of version 0.1 of the gui. User feedback will be helpful and the best feedback will come after folks have tried out the new gui.

CharliePoole:
This fixes #48.

* Right-click now selects the node that is clicked on. This seems more usable, since it shows which node will be affected by the context menu. It also results in displaying the properties for the selected test, which is handy after a run.

* This implementation uses the ids of the selected tests. For certain groups, it would be more convenient to use a category filter, but that will require additional work in order to combine id and category filters. This can be dealt with separately at a lower priority.


CharliePoole:
@nunit/core-team @nunit/contributors This PR has been up for a few days without comment so I'm going to merge it. I would definitely like to get people reviewing the gui stuff, but at this stage I guess it's OK to keep pushing out changes so people can see what the gui looks like.

That said, if you would like to participate, please ensure you are getting notices from the nunit-gui project. I'm going to merge this PR as well as #81.

PR #83, opened 2016-01-25
Fixes #68

### LINKED ISSUE #68: Implement XML view of test details.
The XML tab of the GUI display is currently empty.

iCodeIT:
Add synax highlighting to the XML tab.

### DIFF
--- a/src/nunit-gui/Views/XmlView.Designer.cs
+++ b/src/nunit-gui/Views/XmlView.Designer.cs
@@ -61,6 +61,7 @@
             // xmlTextBox
             // 
             this.xmlTextBox.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.xmlTextBox.Font = new System.Drawing.Font("Courier New", 10F, System.Drawing.FontStyle.Regular, System.Drawing.GraphicsUnit.Point, ((byte)(0)));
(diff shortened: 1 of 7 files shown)


iCodeIT:
I decided to not add word wrap - it looks kind of silly when the xml is formattet nicely with indentations. 
Syntax highlighting colors are inspired the xml-viewer in Google Chrome.

CharliePoole:
This looks great! Two questions...

1. Did any of the code come from another app? If so I have to review the license.

2. I think some people would like to have the option of word-wrap. What about having a context menu with copy to clipboard and word-wrap on it? If you want, I can write an issue and you could do that separately.

I have noticed that there is a larger unrelated issue in the content, which is that the test xml is shown without the result of a run, even if availble. That's definitely for a separate issue though.

iCodeIT:
1. I looked at some code on the net and an rtf specification, and then wrote the code from scratch. I actually thought about writing a blog post on how to do it. Maybe put the code on Github with a license of 'use as you like' 

2. Please do

3-ish. I would get that from the result node, right? Please create an issue for this.

CharliePoole:
That's great - I just wanted to be sure. 

I'll create issues for both of the extra items and assign to you. They are both fairly small.

PR #87, opened 2016-01-27
Fixes #84

### LINKED ISSUE #84: Add context menu to XmlView
Suggested menu items:
 * Select All
 * Copy
 * Word Wrap (checked item)

iCodeIT:
Context menu added to xml view. 
Support for Select All, Copy and Word Wrap.


CharliePoole:
General comments:

* Select All doesn't show the selection all the time. Most times it appears that there is no selection. You probably need to refresh the display after selecting all.

* Copy should be disabled if nothing has been selected either manually or via select all.

src/nunit-gui/Views/XmlView.cs
```diff
@@ -45,6 +45,9 @@ public XmlView()
             InitializeComponent();
 
             XmlPanel = new ControlElement<Panel>(xmlPanel);
+            selectAllToolStripMenuItem.Click += (s, a) => xmlTextBox.SelectAll();
+            copyToolStripMenuItem.Click += (s, a) => xmlTextBox.Copy();
+            wordWrapToolStripMenuItem.CheckedChanged += (s, a) => xmlTextBox.WordWrap = wordWrapToolStripMenuItem.Checked;
```
CharliePoole: Just noting that you are not going through the presenter here. If we wanted to test the behavior, we would have to feed the events to the presenter through some kind of ViewElement. Then NSubstitute would work for us in mocking the view. Don't change it now, however, since we may change the architecture before we are done.

iCodeIT: I was a bit uncertain what to do here, since it is pure gui-stuff happening here. I don't mind moving it.

iCodeIT:
I didn't experience the SelectAll problem when I tested it last night - I do see it now.
I will fix those issues and get back to you.

iCodeIT:
Code is updated. Sorry it took so long.

CharliePoole:
This looks good and I'm merging it.

Regarding my inline comment...

You aren't strictly following the architecture as it is used in the other views. That's OK since I haven't documented it and I'm not sure we will retain it. But just so you know, what I have tried to do is...

* Always go through the presenter. Even view events that merely lead to a view reaction (e.g. Copy) is handled through the presenter.

* Any gui widgets that have to be dealt with externally are wrapped by an Element of some kind and implement an interface.

* We use reflection in the view tests to find any elements in the view and ensure that they are properly initialized.

The purpose of all the above is to make everything testable by ensure that the view interface is mockable using NSubstitute. NSubstitute works best with interfaces.

It may be good for you to look through some code for one of the other view/presenter combos to see what I mean. ProgressBar and StatusBar are pretty good starting points.

At the moment, my priority is to get a version of the gui released for usability testing. Even if all features don't work, I want people to comment on what it looks like and give us some feedback. Following on that, I would like to have a structured code review with other team members - first the model, then the views, etc. - and finally to get back to work on the gui using the same review and testing standards we use for nunit itself.

iCodeIT:
If I understand you correctly, I should have made an IRichTextBoxElement in the UiKit? And expanded the IToolStripItem with a Checked property.
And the actual updating of gui-elements happens in the view by calling interfacemethods from the presenter. In the view we use InvokeIfRequired (as we should). 
Is that fairly accurate?

CharliePoole:
That would have been consistent with what we've done up to now. We already have a checked property for menu elements. We have not tried to use coloring before, so a new or enhanced text element would be needed.

The view is kept very dumb, not even including what I sometimes call "gui logic." For example, Copy would trigger an event and the presenter would actually get the selected text and copy it to the clipboard. Generally, InvokeRequired is called in the element, rather than in the view code.

This is classic MVP using the Presenter First paradigm, with the added abstraction layer of Elements, which comes from my older style of using custom controls. Elements provide an interface that our mocking framework knows how to implement, handles the Invoke if needed and in a few places provides some enhancement to the underlying control. IOW, where I used to create a derived control, I sometimes now place the enhanced logic in the Element. TreeElement is an example of this, while ProgressBar uses the more traditional inheritance approach.

I don't consider this architecture as final. It would be useful for you to try to get some experience with it in order to be able to participate in the discussion about whether to keep it, but you probably don't want to go overboard with it unless we decide to keep it. If you see some place in the code where you might like to try to follow this pattern, let me know.

PR #88, opened 2016-02-04
Fixes #85

### LINKED ISSUE #85: Display test result info in XmlView
When a test has been run, the xml displayed should be that for the result rather than simply for the test exploration. The model method GetResultForTest will return it if available.

One tricky bit: the results are isolated - that is the child results are not included but must be fetched separately. This may make it harder than I originally imagined, in which case it can be delayed.

iCodeIT:
XMl view now shows XML for resultnodes if available


iCodeIT:
Closing this one, as a new (identical) one has been made on nunit-gui

PR #89, opened 2016-02-04
Fixes #85

### LINKED ISSUE #85: Display test result info in XmlView

iCodeIT:
Show both result nodes and test nodes in the xml view.

### DIFF
--- a/src/nunit-gui/Presenters/XmlPresenter.cs
+++ b/src/nunit-gui/Presenters/XmlPresenter.cs
@@ -22,9 +22,12 @@
 // ***********************************************************************
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Text;
 using System.Xml;
+
+using Mono.Cecil;
 
 namespace NUnit.Gui.Presenters
 {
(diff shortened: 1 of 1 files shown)


iCodeIT:
Ping @CharliePoole ;-)

src/nunit-gui/Presenters/XmlPresenter.cs
```diff
@@ -99,5 +99,31 @@ private void DisplayXml()
...
+            if (resultNode != null)
+            {
+                currentXml = resultNode.Xml.Clone();
+            }
+            else
+            {
+                currentXml = testNode.Xml.Clone();
+                while (currentXml.ChildNodes.Count > 0)
+                {
+                    currentXml.RemoveChild(currentXml.ChildNodes[0]);
```
CharliePoole: Here, I believe you are removing child nodes that should not be removed: properties for example, possibly element content. I think you need to have a method that removes child tests only, that is, elements with the name "test-case" or "test-suite".

CharliePoole:
Looks generally right, with the exception of my comment above. Once you fix that I'll pull it down and try it - nothing like actually seeing the gui. :-)

iCodeIT:
@CharliePoole  Ok, I have made the changes to the way xml-nodes are replaced. I think this should work - at least I see both property nodes and CDATA nodes now.

CharliePoole:
Looks good to me... I'm merging.

For next time... our naming standard for branches in our own repos is like issue-85 rather than Issue85. We don't do anything automated with the gui branches yet, but we may in future.

PR #101, opened 2016-05-12
Bug Fix when viewing by category

Poulisj:
When viewing testcases by category, only the first test of each case gets listed under the correct category, everything else defaults to Category None (via the groups.Add(Groups[0]) line above), even if an existing group was found with the required name.

### DIFF
--- a/src/nunit-gui/Presenters/CategoryGrouping.cs
+++ b/src/nunit-gui/Presenters/CategoryGrouping.cs
@@ -92,8 +92,9 @@ namespace NUnit.Gui.Presenters
                 {
                     group = new TestGroup(groupName);
                     Groups.Add(group);
-                    groups.Add(group);
                 }
+                
+                groups.Add(group);
             }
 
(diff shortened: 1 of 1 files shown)


iCodeIT:
I think you should have check for not adding the samme group to groups twice, unless we are absolutely sure it can't happen? @CharliePoole ?

Otherwise it seems like a good fix

CharliePoole:
It looks good to me as well. In principle, we will only add it once because once we add it it should always be found.

I'd feel better if we had a test for this but I'm OK with manual verification for now anyway.

iCodeIT:
Ok - I will merge this.

CharliePoole:
@iCodeIT When merging a PR that doesn't reference an issue, we label it just like an issue so that it will show up for listing in our readme file. Since no next milestone is defined, I just assigned this to backlog. We'll move done stuff from backlog to the next release milestone once it's created.

PR #107, opened 2016-09-11
Views

immeraufdemhund:
Small change to structure of code by adding a new project for IViews. What I'm aiming to accomplish is to make the views have next to no logic in them. With that said I made a small change to XmlView. I believe this moves view concerns back into the view while not affecting overall functionality. I'd like to keep down this path of simplifying the Views.


CharliePoole:
The problem with the way the change is presented is that it's a list of 80 files that changed or are new. That means two committers have to examine all 80 files to figure out what you have done. Just looking at a small subset of files, I can see that many of the changes have nothing whatsoever to do with wht you are trying to accomplish. You have touched various files that are only changed for minor reasons, such as eliminating unused using statements. Many of those independent changes are good changes, but they muddy the waters for someone who has to understand your code.

Based on the description, I guess your main objective here is to have a slightly different architecture for nunit-gui. You want to keep the MVP structure, but have a separate interface for the views in order to add future view implementations. I'd prefer that we have discussed that first, since it isn't really an objective of the project as it now stands to be anything but a Windows Forms application.

Without making an advance judgement on your re-design, could we back off for a moment and discuss what alternative implementations you are thinking of and how an MVP architecture might work in those implementations?

immeraufdemhund:
>Many of those independent changes are good changes, but they muddy the waters for someone who has to understand your code.

I appreciate your concern with the many file changes. I get dinged for that at work too (one of these days I'll break that habit). I could do some git magic and make a seperate pull request JUST for the automated cleanup stuff if you'd like. As for the slight restructure what I was hoping for was to make it have different UI technologies so that a person could relatively easily workout a test explorer for visual studio (I'd want to contribute work to that)

> Based on the description, I guess your main objective here is to have a slightly different architecture for nunit-gui. You want to keep the MVP structure, but have a separate interface for the views in order to add future view implementations

In my mind I was just moving what you already had in place but just to a new project. Along the way I did get distracted by trying to get the winforms and drawing stuff out of the project (which is why the XmlView got a minor tweaking) so I (probably) should have focused more on just moving than cleaning.

>Without making an advance judgement on your re-design, could we back off for a moment and discuss what alternative implementations you are thinking of and how an MVP architecture might work in those implementations?

I'm fine with sticking with MVP since the logic is already in place and making a WPF project (I think that is the underlying tech for a visual studio plugin) use MVP sticks with the separation of concerns (granted all the WPF MVVM purists would have my head for that, but they wouldn't think twice about just dumping the winforms app and starting fresh)

CharliePoole:
No need to redo it. But for future, small, targeted issues and PRs are the way to go. Also, issues for discussion prior to doing the PR is often a way to avoid wasting your own time.

I'll go through this in the next day and focus down to the essentials for review.

immeraufdemhund:
Can we hold off on it? I don't mind creating a separate branch with cleanup stuff only. After you saying what you did I feel like we should collaborate a little and either I pick up a different story here in your issues and work on that. With that in mind how come the project uses .net 2.0? Isn't mono using 4+ and with windows xp no longer being supported that gives the windows user 4.5+? Lastly can we use the latest NUnit engine with this?

CharliePoole:
Sure... if you prefer, that would be great. I just don't want to discourage anyone who puts in the time to work on the project.

We target .NET 2.0 because that's our minimum target platform. Of course, it runs on higher versions of the framework. I considered using a 4.x version for the GUI, but ended up not doing it because there seemed to be no drawback. If something turns up, we can talk about it. Note that the engine also  targets 2.0, and that's not something we want to change.

One thing to remember is that people set up legacy systems for testing all the time if they want to be able to run on those platforms. It's systems like those that we are thinking of when we set the level, not the typical developer desktop machine.

immeraufdemhund:
I was hoping to consider the use of the TPL in .net 4.0. I am going over the idea of having some of the heavy lifting done in the views work in the background and I am just more familiar/comfortable with Task than I am background worker. (Not that I wouldn't have to google my way through it using either because I work primarily with background services.)

CharliePoole:
How come you're closing this?

immeraufdemhund:
I read over the documentation about how developers should act/code on this project and I felt that I had too much automated clean up and big changes and so forth. I would still like to go down that path still but I dunno do it in stages or maybe one thing at a time. maybe I could actually start by creating the single project and move over IView, IDialog, and my new IAddinsView to it.

________________________________

CharliePoole:
@immeraufdemhund Please excuse the comment here - I have no other way to contact you.

You requested to join the Gui Team on GitHub and I denied it. I expected GitHub to pop up a box that let me add a message to you, but it didn't.

Actually, I didn't even know you could request to join a team that way! 

Generally people are invited to the Contributors Team and work that way for a while before being invited to one of the other teams. Since we only just added you as a contributor, it seems a bit soon to be making you a committer on the GUI.

I hope you'll continue to work on the project so we can see what you are able to do.

PR #109, opened 2016-09-13
updated project to use latest NUnit.Engine and Framework.

immeraufdemhund:
issue#108

this does not fix the issue stated, but it is where the idea was planted in my head to update (granted we talked about it in a differen thead)

This pull request simply is just to update to the latest version of NUnit Framework and Engine. However I feel like maybe a slight discussion should take place about this because I feel like a better way to do this would be to use a submodule. Then I could point to specific commits in there if say I want to pull in a fix. I'm not sure how the build server would like it though. I suppose one way to find out.

### DIFF
--- a/src/tests/Model/TestModelTests.cs
+++ b/src/tests/Model/TestModelTests.cs
@@ -21,12 +21,14 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-using NUnit.Engine;
-using NUnit.Framework;
-using NUnit.Gui.Tests.Assemblies;
+using System.IO;
 
 namespace NUnit.Gui.Model
 {
+    using Engine;
+    using Framework;
+    using Tests.Assemblies;
+
     [Explicit("Second TestEngine conflicts with main engine")]
     public class TestModelTests
     {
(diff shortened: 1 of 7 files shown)


src/tests/Model/TestModelTests.cs
```diff
@@ -37,12 +38,14 @@ public class TestModelTests
         [OneTimeSetUp]
         public void CreateTestModel()
         {
-            var engine = TestEngineActivator.CreateInstance();
-            Assert.NotNull(engine, "Unable to create engine instance for testing");
+            using (var engine = TestEngineActivator.CreateInstance())
```
CharliePoole: This looks like a potential bug. We create an engine and give it to the model. The model continues to exist after the method exits and is used by the rest of the tests. It holds and uses the engine, which has already been Disposed on exit from the using. 

Apparently, we don't use anything in the current tests that gets disposed, or they wouldn't be passing. But we shouldn't Dispose the engine directly at all, since the model controls it's lifetime.

CharliePoole:
Looks good except for the comment about Disposing the engine, which should be fixed.

immeraufdemhund:
amended my last commit. Please see commit de2e104bdfb42235b0944d21412ce79797e5d631 (it should have auto updated, but just in case it didn't)

PR #110, opened 2016-09-13
removed NuGet version of NUnit-Engine in favor of using a submodule.

immeraufdemhund:
issue#108

not sure how the build server is going to react to a submodule. Probably would just need to be told to get said submodule if it fails in the build steps.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "nunit-console"]
+	path = nunit-console
+	url = https://github.com/nunit/nunit-console.git
--- a/nunit-gui.sln
+++ b/nunit-gui.sln
@@ -23,6 +23,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		LICENSE = LICENSE
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "nunit-engine", "nunit-engine", "{3F520B74-13C3-45E6-A3A0-472A1968F2F1}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "nunit.engine.api", "nunit-console\src\NUnitEngine\nunit.engine.api\nunit.engine.api.csproj", "{775FAD50-3623-4922-97C4-DFB29A8BE4C7}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "nunit.engine", "nunit-console\src\NUnitEngine\nunit.engine\nunit.engine.csproj", "{372A3447-D657-40FF-A089-77C19FEC30C8}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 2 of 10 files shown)


immeraufdemhund:
looks like AppVeyor is missing the `git submodule update --init` call like the Travis CI build is doing.

CharliePoole:
@immeraufdemhund This isn't really something we want to do. Probably the most fundamental change in NUnit 3, was dividing NUnit into three independent layers. As it stands, the separation isn't yet complete - we just split the framework from the other two and will be doing independent releases in the future. The same thing is planned for the engine, so tying the engine build together with the Gui would be a step backwards from where we are trying to go.

There is no doubt that the need to wait for new releases of the engine can sometimes be painful, especially as the gui tends to make new demands on the engine that the console runner didn't call for. But we think it will be worth the effort in the end and will enable multiple runners to use the same engine, in many cases without needing to install it.

immeraufdemhund:
if I may interject about your separation concerns. (Please realize this might just be my misunderstanding of what a submodule is). With a submodule it is just a reference to a specific commit in a repo. You could (but shouldn't) make changes in the submodule and commit those back to the submodule but the code base and changes would only point to the submodule. Not to mention that people can work on that submodule without any restrictions and commit as much as they want and I wouldn't be affected because I would still be pointing at a specific point in time. If I wanted I could update (for a bug fix, or a new feature..etc). But in the end it is still (essentially) two git repositories.

for instance if you look here at the source tree from my remote: [https://github.com/immeraufdemhund/nunit-gui/tree/issue-108-submodule](url) you'll see that the nunit-console source is not in nunit-gui at all. There is a new folder which will go to your nunit-console at a specific commit point.

Or another way I like to think of it is like NuGet in that I ask for a specific version but I also get the source code locally for debugging purposes.

CharliePoole:
@immeraufdemhund Yes, I understand what a submodule is, of course. :-)

We've considered using them for some other purposes (source-level-shared code like command-line option processing, for example) but haven't done it yet.

Adding it in temporarily as a way to debug is an interesting idea I hadn't thought of. I have generally accomplished the same thing by temporarily adding the local engine project to the gui solution. This seems to work equally well and has the advantage that I'm only pointing at my local copy of the gui.

Are there any ways to use a submodule, while restricting how it's used to read-only? That might be something to consider.

But the final objection to replacing packages with a submodule is that we distribute NUnit in discrete packages - getting more and more discrete with each release, in fact. In principle, the GUI (and any runner) is supposed to work with whatever engine you have already installed.

You might say we are in a phase of unbundling things, which makes a snapshot taken today look either half-bundled or half-unbundled, depending on how you look at it. :-)

@rprouse What do you think about the possibility of using submodules as @immeraufdemhund  describes? Either for this purpose or for some others. One small way I have considered using a submodule is to divide our cake scripts into useful sub-scripts and share some of them that way. However, the update issue is a blocker for me: I don't want folks updating things that will have an effect on multiple projects without testing those projects.

immeraufdemhund:
@CharliePoole  said
>However, the update issue is a blocker for me: I don't want folks updating things that will have an effect on multiple projects without testing those projects.

(Sorry if I'm stepping on toes, still learning here) Wouldn't each project be responsible for testing the application if they decide to update their submodule.  Here at my job we use them rather frequently and we only update submodules when we have a need. And if another concern is that the behavior is going to change a person could write a test more as documentation of how they use a project (stored in the submodule) so that someone knows if their change is breaking a different projects use. Or the inverse could be that you would write tests around how you use a project and then if you decide to update your submodule you'll see if something breaks.

CharliePoole:
I think what you may be missing is that the engine is a separate product in itself, distributed separately. We expect every runner to work with the latest release of the engine.

The change you are suggesting is a way that we could work __if__ the engine were merely a library used by other runners and if we wanted to allow each runner to have it's own different version of the engine. But that's not the goal we are aiming for.

The idea is that the GUI package __will not include an engine__ but will use the version of the engine it finds on the machine where it is installed. We would like to do the same thing for each runner, although we will also offer "full" packages for folks who don't already have a runner installed.

I should mention that this has some benefit for our own runners, but is of even greater potential benefit to third-party runners.

PR #118, opened 2016-10-12
When initializing ResultNode use CultureInfo.InvariantCulture for parsing `duration` attribute value

### LINKED ISSUE #117: System.FormatException on parsing duration
A 'System.FormatException' Exception is raised on parsing duration attribute.

Class: ResultNode
Method: InitializeResultProperties
Line 56: Duration = (duration != null) ? **double.Parse(duration)** : 0.0; 

Proposal:
adding "System.Globalization.CultureInfo.InvariantCulture" into the parse call.
->double.Parse(duration , System.Globalization.CultureInfo.InvariantCulture)

romansp:
Fixes #117.

Inside `InitializeResultProperties` changed to `double.Parse` overload with culture parameter usage, so it will parse using `CultureInfo.InvariantCulture` and thus won't throw `System.FormatException` on systems that use different from '.' (dot) decimal delimiter in their culture, e.g. ',' (comma).

### DIFF
--- a/src/nunit-gui/Model/ResultNode.cs
+++ b/src/nunit-gui/Model/ResultNode.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Text;
 using System.Xml;
 using NUnit.Engine;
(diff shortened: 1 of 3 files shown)


CharliePoole:
Looks good. Thanks!

PR #136, opened 2016-12-11
Eliminate console window display when running in separate process

CharliePoole:
This fix makes use of a pre-release build of NUnit Engine 3.6 to avoid creating a window when launching the agent process. It also fixes a problem that prevented use of RunAsync in the gui when running tests.

### DIFF
--- a/src/nunit-gui/nunit-gui.csproj
+++ b/src/nunit-gui/nunit-gui.csproj
@@ -34,15 +34,15 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Mono.Cecil, Version=0.9.6.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.0-ci-03382-pr-146\lib\Mono.Cecil.dll</HintPath>
+      <HintPath>..\..\packages\NUnit.Engine.3.6.0-dev-03383\lib\Mono.Cecil.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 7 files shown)


src/nunit-gui/nunit-gui.csproj
```diff
@@ -34,15 +34,15 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Mono.Cecil, Version=0.9.6.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.5.0\lib\Mono.Cecil.dll</HintPath>
+      <HintPath>..\..\packages\NUnit.Engine.3.6.0-ci-03382-pr-146\lib\Mono.Cecil.dll</HintPath>
```
immeraufdemhund: @CharliePoole is there a way in nuget settings to make these paths static? if not no worries.

CharliePoole:
You don't want them to be absolute. The projects are downloaded by each developer to a potentially different place on their disk drive

Since @rprouse just merged my fix to the engine into master, I'm going to change the reference form a `-ci-` to a `-dev-` build before I merge this.

immeraufdemhund:
What I meant was more of a relative absolute. instead of `..\..\packages\NUnit.Engine.3.6.0-ci-03382-pr-146` `..\..\packages\Nunit.Engine_HEAD` or something like that. But that would make it hard on NuGet because it does a lot of cross checking of assemblies and versions and dependencies and so forth.

CharliePoole:
I see... In general, we deliberately avoid any automatic updating of dependencies to the latest versions, preferring to create an issue and PR whenever we need to do it. That keeps things more independent than they would otherwise be. Remember too that making these different components independent of one another was one of our goals for NUnit 3.

PR #138, opened 2016-12-12
Remove unused settings panels and settings

CharliePoole:
Continued work on issue #86.


src/nunit-gui/Views/SettingsPages/GuiSettingsPage.cs
```diff
@@ -28,11 +28,11 @@
 using System.Data;
 using System.Text;
 using System.Windows.Forms;
+using NUnit.Gui.Model;
```
immeraufdemhund: In the coding standards you mentioned putting NUnit usings inside the namespace. Are we moving away from that? (I ask that because you moved the Model.Settings up here.

CharliePoole: No, the standard says "permissible but not required" I ended up with Model in one place and Model.Settings in another and it looked funny to me so I moved them both to the outside.

PR #141, opened 2016-12-16
Fix errors running in separate process

### LINKED ISSUE #61: When loading tests using an agent, a cmd window is opened
One window is opened for each nunit project when using the file open menu to open an assembly out of process, using nunit-agent~~and selecting a project file such as an nunit project~~. Nothing is displayed in the window and it can be closed immediately, but it's a distraction.

### LINKED ISSUE #133: Initial NUnit Tree display should show top of tree
When the NUnit Tree display is selected, the tree is initially shown at a random location. We should show the top of the tree instead.

### LINKED ISSUE #134: File Exit command doesn't work
The only way to close the Gui currently is to click "X" in the upper right corner.

CharliePoole:
Fixes #134 Fixes #61 Fixes #133 

Several small errors, as detailed in the issues

### DIFF
--- a/src/nunit-gui/Model/TestModel.cs
+++ b/src/nunit-gui/Model/TestModel.cs
@@ -187,7 +187,11 @@ namespace NUnit.Gui.Model
 
         public void Dispose()
         {
-            throw new NotImplementedException();
+            if (IsPackageLoaded)
+                UnloadTests();
+
+            if (Runner != null)
+                Runner.Dispose();
         }
 
         public void NewProject()
--- a/src/nunit-gui/Presenters/NUnitTreeDisplayStrategy.cs
+++ b/src/nunit-gui/Presenters/NUnitTreeDisplayStrategy.cs
@@ -70,8 +70,7 @@ namespace NUnit.Gui.Presenters
                 SetInitialExpansion(treeNode);
             }
 
-            if (topNode != null)
-                topNode.EnsureVisible();
+            topNode?.EnsureVisible();
         }
 
         private void SetInitialExpansion(TreeNode treeNode)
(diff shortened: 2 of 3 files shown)


src/nunit-gui/Model/TestModel.cs
```diff
@@ -187,7 +187,11 @@ public void OnStartup()
 
         public void Dispose()
         {
-            throw new NotImplementedException();
+            if (IsPackageLoaded)
+                UnloadTests();
+
+            if (Runner != null)
```
immeraufdemhund: now that we can use C# 6.0 features this could change to Runner?.Dispose();

src/nunit-gui/Presenters/NUnitTreeDisplayStrategy.cs
```diff
@@ -57,14 +57,21 @@ public override void OnTestLoaded(TestNode testNode)
...
+                if (topNode == null)
+                    topNode = treeNode;
+
                 _view.Tree.Add(treeNode);
 
                 SetInitialExpansion(treeNode);
             }
+
+            if (topNode != null)
+                topNode.EnsureVisible();
```
immeraufdemhund: same here with the null propogation. topnode?.EnsureVisible();

CharliePoole: It's early days, but I'll give it a try. We really need to formulate some sort of standard about how agressively we want to use this and other features.

src/nunit-gui/Model/TestModel.cs
CharliePoole: In this case, I don't think so. It's common in our Dispose and some other cleanup methods to have a sequence of `if` statements each governing some part of the disposal. In this case, the first one can't be changed to use the `?.` operator, while the second one can. That would leave us with

```C#
if (IsPackageLoaded)
    UnloadTests();

Runner?.Dispose();
```

This breaks the pattern and requires a bit more thought to see that it's a sequence of conditionals.

immeraufdemhund:
merged

PR #149, opened 2016-12-25
Revert "Remove additional unused settings"

immeraufdemhund:
Reverts nunit/nunit-gui#146


CharliePoole:
@immeraufdemhund Thanks for doing this, but I'm going to leave the change in the gui. I'm moving ahead with the engine changes that match it.

PR #152, opened 2017-01-03
Issue 75

### LINKED ISSUE #75: Add a summary display to the GUI
We need some place in the gui that displays tests run, not-run, etc. as we do in the V2 GUI. Ideally, it should list every different result type.

Best would be to have a dedicated panel for this purpose. If it gets too large for that, we could have a separate summary tab.

immeraufdemhund:
resolves #75


CharliePoole:
I think we'll want to do more on this display, but I'm merging what you have so it can be more easily seen and commented upon. Thanks for the work!

PR #153, opened 2017-01-09
Typo fixes

jnm2:
Continuing nunit/nunit#1960, just a couple changes.

[One is in some dead code. Can I remove it?](https://github.com/nunit/nunit-gui/pull/153/files#r95192666)

/cc @CharliePoole @ChrisMaddock

### DIFF
--- a/src/nunit-gui/CommandLineOptions.cs
+++ b/src/nunit-gui/CommandLineOptions.cs
@@ -190,7 +190,7 @@ namespace NUnit.Gui
                 if (int.TryParse(val, out r))
                     result = r;
                 else
-                    ErrorMessages.Add("An int value was exprected for option '{0}' but a value of '{1}' was used");
+                    ErrorMessages.Add("An int value was expected for option '{0}' but a value of '{1}' was used");
             }
                 
             return result;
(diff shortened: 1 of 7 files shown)


src/nunit-gui/CommandLineOptions.cs
```diff
@@ -190,7 +190,7 @@ private int RequiredInt(string val, string option)
                 if (int.TryParse(val, out r))
                     result = r;
                 else
-                    ErrorMessages.Add("An int value was exprected for option '{0}' but a value of '{1}' was used");
+                    ErrorMessages.Add("An int value was expected for option '{0}' but a value of '{1}' was used");
```
jnm2: This is in a method which is not used. Underneath is commented-out code. Should I remove the whole method, possibly also the commented-out code?

CharliePoole:
This looks good. I'll merge when the CI completes.

Frankly, however, I think that this cross-cutting approach to fixing typos is a poor use of our time. But if you're going to do it, you should __only__ fix typos in the PR. We can rapidly review that they are just typos and get it merged. If you start making other changes, we need to analyze the code to figure out if they are OK. This applies even to something as "obvious" as a method that is not called (any longer? yet? is somebody working in this code? Is it shared across projects?) This is especially true in a non-production, experimental project only worked on by a few people - like this one.

jnm2:
> Frankly, however, I think that this cross-cutting approach to fixing typos is a poor use of our time.

It makes @ChrisMaddock's and my inner perfectionists happy =) I did one, so I just did all at the same time.

CharliePoole:
Like I said, so long as it's kept separate from changes that require thinking, I'm cool with it.

PR #157, opened 2017-01-14
#156 Update to RunTree image

### LINKED ISSUE #156: Fix RunTree image for transparency.
The RunTree image currently has a white background instead of a transparent one.

JustinRChou:
Resolves #156
Update to RunTree image to a transparent background


CharliePoole:
The image looks good but the resx file is a bit messed up. The linux error is because you didn't use the right casing for the file name. But the original resx didn't have a filename at all because the bitmap was simply embedded. That got changed to a file reference. In addition, the resources you didn't change got changed in the file. I'm wondering if it's because you use a different version of VS. I've been using VS2015 community for this project. Ideally, it shouldn't matter of course.

JustinRChou:
I working in VS2015.
I did drag and drop the file into the resx and just deleted the old one.

Going to a readd.

CharliePoole:
Much better. Can you squash so we don't have the add / revert / readd ? Alternatively, reset to before you started, recommit and do a forced push.

JustinRChou:
I would prefer to squash merge, but I have no write access here.

CharliePoole:
In your local branch you can do 

```
git reset <commit> // probably HEAD~3
git add .
git commit -m "Update Run Tree image"
git push -f
```

and the PR will be updated. Otherwise, I'll pull it down and do it.

JustinRChou:
Neat.

CharliePoole:
Merging... thanks for the work!

PR #159, opened 2017-01-17
SettingsLoaded only true after LoadSettings call

### LINKED ISSUE #130: Corrupted NUNit30Settings file
When changing the number of recent files setting, the NUnit30Settings file becomes corrupted. I observed this when reducing the number of settings from 5 to 3, when there were already 3 files shown in the list.

mikkelbu:
Otherwise, we could end up in a situation where ApplySettings
is called even if LoadSettings is not called.

Fixes #130

### DIFF
--- a/src/nunit-gui/Views/SettingsPage.cs
+++ b/src/nunit-gui/Views/SettingsPage.cs
@@ -22,11 +22,6 @@
 // ***********************************************************************
 
 using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Drawing;
-using System.Data;
-using System.Text;
 using System.Windows.Forms;
 
 namespace NUnit.Gui.Views
@@ -56,16 +51,15 @@ namespace NUnit.Gui.Views
 
         #region Properties
 
+        private bool settingsLoaded;
+
         protected SettingsModel Settings { get; private set; }
 
         protected IMessageDisplay MessageDisplay { get; private set; }
 
         public string Key { get; private set; }
 
-        public bool SettingsLoaded
-        {
-            get { return true; }// _settings != null; }
-        }
+        public bool SettingsLoaded => settingsLoaded;
 
         public virtual bool HasChangesRequiringReload
         {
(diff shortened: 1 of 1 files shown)


src/nunit-gui/Views/SettingsPage.cs
```diff
@@ -86,13 +80,19 @@ public virtual void ApplySettings()
 
         #endregion
 
+        void LoadAndSetSettings()
```
CharliePoole: Why two methods? When do you want settings to be loaded without setting the flag?

src/nunit-gui/Views/SettingsPage.cs
```diff
@@ -56,16 +51,15 @@ public SettingsPage(string key, SettingsModel settings)
...
 
         protected IMessageDisplay MessageDisplay { get; private set; }
 
         public string Key { get; private set; }
 
-        public bool SettingsLoaded
-        {
-            get { return true; }// _settings != null; }
-        }
+        public bool SettingsLoaded => settingsLoaded;
```
CharliePoole: From the comment on the old code, it looks as if the property formerly checked that _settings is not null. I can't remember why we stopped doing that. Usually, when something is commented out like that, it means I'm in the middle of working on it and made a temporary change. The change should not be in master, but it may have been committed by mistake.

mikkelbu: I've changed `SettingsLoaded ` into an autoproperty. If `_settings_` could be `null` then I think that we will get NRE in `SettingsDialog`.

src/nunit-gui/Views/SettingsPage.cs
mikkelbu: An error on my part. I wanted to make sure that the flag was always set when overriding `LoadSettings`. `SettingsLoaded` is now set in `OnLoad`.

mikkelbu:
I don't think that it is the responsibility of the model to know whether some properties of the model has been used to update the view.

Perhaps the methods `LoadSettings` and `ApplySettings` (and the property `SettingsLoaded`) should be renamed to make it more explicit that we are just updating the view/updating the model, respectively.

mikkelbu:
@CharliePoole I made some changes and a comment after the review, but I'm unsure whether github notified you about this?

CharliePoole:
Sorry this has sat for so long - I haven't been paying attention to the gui. 😞

PR #168, opened 2017-03-17
Honor "Initial Tree Display" setting

mikkelbu:
The "Initial Tree Display" setting is now honored. For 'Auto'
we mimick the behaviour of the nunit 2 gui and expand all
the nodes, if the the number of nodes are below the number
of visible nodes; otherwise we collapse to fixtures.

Fixes: #131


CharliePoole:
@mikkelbu Did you ever figure out why this was failing in the CI?

mikkelbu:
@CharliePoole  I think I forgot to run tests before making the PR (I always forget that running tests are not the default). Furthermore, I misread the error message, so I thought that the error was a temporary fluke on the BS.
The problem is that `_model.Settings` is null in the test. I think I will solve this by either set `Settings` to something meaningful in the test, or by making the production code more robust. I'll probably have time to look into this in a couple of days.

CharliePoole:
@mikkelbu There was a fix (yours I think) that corrected a problem with detecting whether settings are loaded. You should probably merge master and retest. This one is the last of our old PRs, which I had been neglecting for so long.

mikkelbu:
@CharliePoole I've added a commit which sets default settings for the tests which in turn makes the test pass

PR #175, opened 2017-04-17
Update Message and StackTrace

### LINKED ISSUE #173: Test result message and stacktrace not always updated
```cs
using System;
using NUnit.Framework;

[TestFixture]
public sealed class Test
{
	[Test]
	public void Inconclusive()
	{
		Assert.Inconclusive("inconclu");
	}

	[Test]
	public void Fail()
	{
		Assert.Fail("failmsg");
	}

	[Test]
	public void Success()
	{
	}
}

```
e.g. Success test is shown green but the stacktrace and message belongs to Fail.
image
Even after re-running Success, the text is not updated.

mikkelbu:
Message and StackTrace are now reset to "" if the
test does not contain a "message"/"stack-trace"
XML node, repectively.

Fixes #173

### DIFF
--- a/src/nunit-gui/Presenters/TestPropertiesPresenter.cs
+++ b/src/nunit-gui/Presenters/TestPropertiesPresenter.cs
@@ -124,11 +124,8 @@ namespace NUnit.Gui.Presenters
                         message = resultNode.Xml.SelectSingleNode("reason/message");
                     }
 
-                    if (message != null)
-                        _view.Message = message.InnerText;
-
-                    if (stackTrace != null)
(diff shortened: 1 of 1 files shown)


CharliePoole:
Looks good to me. Thank you!

@immeraufdemhund As small as this is, it's OK that you merged it, but keep in mind that our standard procedure is that it takes two committers to approve a merge from a non-committer - at least until we change it :smile:

immeraufdemhund:
@CharliePoole oops. Sorry about that. I forgot that it was 2.

PR #177, opened 2017-04-25
Upgrade to framework and engine versions 3.6.1

### LINKED ISSUE #176: Upgrade to version 3.6.1 of engine

CharliePoole:
Fixes #176

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -36,6 +36,7 @@ var GUI_TESTS = BIN_DIR + "nunit-gui.tests.dll";
 // Package sources for nuget restore
 var PACKAGE_SOURCE = new string[]
 {
+    "https://www.nuget.org/api/v2",
 	"https://www.myget.org/F/nunit-gui-team/api/v3/index.json",
 	"https://www.myget.org/F/nunit-gui-team/api/v2"
 };
--- a/src/nunit-gui/nunit-gui.csproj
+++ b/src/nunit-gui/nunit-gui.csproj
@@ -34,15 +34,23 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Mono.Cecil, Version=0.9.6.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.0-dev-03383\lib\Mono.Cecil.dll</HintPath>
+      <HintPath>..\..\packages\NUnit.Engine.3.6.1\lib\Mono.Cecil.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="nunit.engine, Version=3.6.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.0-dev-03383\lib\nunit.engine.dll</HintPath>
+    <Reference Include="nunit-agent, Version=3.6.1.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\NUnit.Engine.3.6.1\lib\nunit-agent.exe</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="nunit-agent-x86, Version=3.6.1.0, Culture=neutral, processorArchitecture=x86">
+      <HintPath>..\..\packages\NUnit.Engine.3.6.1\lib\nunit-agent-x86.exe</HintPath>
+      <Private>True</Private>
+    </Reference>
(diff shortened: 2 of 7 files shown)


src/nunit-gui/nunit-gui.csproj
```diff
@@ -34,15 +34,23 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Mono.Cecil, Version=0.9.6.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.0-dev-03383\lib\Mono.Cecil.dll</HintPath>
+      <HintPath>..\..\packages\NUnit.Engine.3.6.1\lib\Mono.Cecil.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="nunit.engine, Version=3.6.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.0-dev-03383\lib\nunit.engine.dll</HintPath>
+    <Reference Include="nunit-agent, Version=3.6.1.0, Culture=neutral, processorArchitecture=MSIL">
```
mikkelbu: Is there any reason for referencing nunit-agent.exe and nunit-agent-x86.exe ?

src/tests/nunit-gui.tests.csproj
```diff
@@ -32,23 +32,31 @@
...
+      <HintPath>..\..\packages\NUnit.Engine.3.6.1\lib\Mono.Cecil.dll</HintPath>
       <Private>True</Private>
     </Reference>
     <Reference Include="NSubstitute, Version=1.10.0.0, Culture=neutral, PublicKeyToken=92dd2e9066daa5ca, processorArchitecture=MSIL">
       <HintPath>..\..\packages\NSubstitute.1.10.0.0\lib\net35\NSubstitute.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="nunit.engine, Version=3.6.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.0-dev-03383\lib\nunit.engine.dll</HintPath>
+    <Reference Include="nunit-agent, Version=3.6.1.0, Culture=neutral, processorArchitecture=MSIL">
```
mikkelbu: Is there any reason for referencing nunit-agent.exe and nunit-agent-x86.exe ?

build.cake
```diff
@@ -36,6 +36,7 @@ var GUI_TESTS = BIN_DIR + "nunit-gui.tests.dll";
 // Package sources for nuget restore
 var PACKAGE_SOURCE = new string[]
 {
+    "https://www.nuget.org/api/v2",
```
mikkelbu: It seems that we are mixing spaces and tabs a lot in this file?

mikkelbu:
@CharliePoole I know that this has already been merged into master, but I think that I have spotted some _very_ minor issues.

Also is there any reason for using nunit v. 3.5 in the cake script and in the tools folder?

src/nunit-gui/nunit-gui.csproj
CharliePoole: That's an error. FYI, the reference is added automatically by nuget and has to be removed manually. Ideally, the agents would be in the tools directory, but the engine expects to find them in the directory where it is installed.

I'll push an update.

build.cake
CharliePoole: This comes up as different people (or the same person at different times) use different editors. We should fix when we are working in the file again but it's not a big priority.

CharliePoole:
I'll take a look at the 3.5 references and update if needed.

src/nunit-gui/nunit-gui.csproj
CharliePoole: This is done for both assemblies. Probably, since I forgot about it, the process needs to be documented somewhere. Given that it's a temporary situation (we hope) do you think it's worth adding a text file in the project root? Or should it go into the wiki docs?

FYI, my plan is that the engine should eventually be in the package tools subdirectory, with only a copy of the api assembly present in the lib directory. That way, nothing else will get referenced by the gui.

build.cake
CharliePoole: Ah! it's also because VS doesn't have any rules for the layout of .cake files. I wonder if we can add a set.

src/nunit-gui/nunit-gui.csproj
CharliePoole: Here's an idea: we can have a readme file in the root, which will be automatically displayed whenever the package is directly installed, but not when it is installed as a dependency. Since the engine package is intended for developers creating runners, it can have lots of developer info.

CharliePoole:
See new PR for fixes.

PR #185, opened 2017-05-19
Implement basic debugging support

### LINKED ISSUE #158: Debugging support
Are you open for a pull request to implement debugging support in the test runner: Starting a test with a debugger attached. Part of running and developing tests is also debugging, and besides Resharper there is not really a suitable graphical test runner available for NUnit 3.

Sebazzz:
Fixes TestCentric/testcentric-gui#235

Implements basic debugging support, using the debugging support from the NUnit Framework itself. Since the debugger is launched from the framework itself, used that mechanism to launch the debugger. This is the most future-proof method, other ways would require modifying the NUnit Framework. If the tests will be lifted out-of-process anytime, this method should still be working.


CharliePoole:
Holding this for the next release.

Sebazzz:
Rebased on latest master.

Sebazzz:
Rebased on latest master.

sidshetye:
@CharliePoole , when will this be merged and released? Is there also documentation on configuring and getting started (e.g. VS project -> properties -> Debug ->Start External Program `nunit-gui.exe`, command line args `project.test.dll` etc) ? I'd like to introduce this to my team. Thanks.

sidshetye:
@CharliePoole This code is waiting for a year to be released ... could you commit either on accepting or rejecting? We'd like a merge obviously but if this is never going to make master, we don't want to be stuck on a custom branch.

CharliePoole:
Team members can somebody review this, please? As you know, I'm limited to my cell phone till the end of the month.

Issues...
1. Is the PR good (obviously)
2. Do we want to take this experiment any further? 
3. Even if we terminate it, should we merge in latest changes before archiving.

@sidshetye Our plan is to review the experiment in June and come up with a future direction for it. Meanwhile, we have been working on the non-experimental gui.

src/nunit.testmodel/Model/TestModel.cs
```diff
@@ -28,12 +28,16 @@
 
 namespace NUnit.Gui.Model
 {
+    using System.Diagnostics;
```
immeraufdemhund: usually we put non-nunit using statements at top of file after header. The ones inside the namespace would be for NUnit usings.

src/nunit-gui/Presenters/TreeViewPresenter.cs
```diff
@@ -97,8 +97,13 @@ private void WireUpEvents()
             };
 
             // View context commands
-            _view.Tree.ContextMenu.Popup += () =>
-                _view.RunCheckedCommand.Visible = _view.Tree.CheckBoxes && _view.Tree.CheckedNodes.Count > 0;
+            _view.Tree.ContextMenu.Popup += delegate
```
immeraufdemhund: Here you use the delegate keyword and below you use a statement labmda. Since this appears to be the only one that uses delegate I suggest changing it to a statement lambda.

src/nunit-gui/Views/TestTreeView.cs
```diff
@@ -62,6 +67,8 @@ public TestTreeView()
 
             RunContextCommand = new MenuElement(this.runMenuItem);
             RunCheckedCommand = new MenuElement(this.runCheckedMenuItem);
+            DebugContextCommand = new MenuElement(this.debugMenuItem);
```
immeraufdemhund: I know you are following suite with using `this.` and there appears to be mixed in here. Can you remove all `this.`s in this file only please.

sidshetye:
@CharliePoole thanks for the update. Excited to see the non-experimental project move forward as it's been a while since I checked in.

My team's primary use case for these GUIs over the VS test runners or CLI `dotnet test` is that we run tests against a dev clusters (service fabric or azure cloud service dev env). Our CI uses `dotnet test` but here is our development cycle use case:

1. Hitting F5 in the VS solution 
2. That starts 2 projects - one that brings up the dev cluster (after several minutes) and another which is the nunit 2 GUI (`nunit.exe test.dll`)
3. When debugging a failure or following TDD, nunit 2 does NOT exit and stop Visual Studio, so we can quickly fire a particular test, trace in the database, edit code in place etc. The turn around is very rapid, usually at pace with the developers. However VS's test explorer or `dotnet test` always exit after the test run, so every code change, every trace run involves the cluster tear down and bring up which frustrates everyone. I mean ~5 minute stalls when folks are 'in the zone' seems frustratingly eternal and that's why the cluster solutions is the only one still sticking to nunit2 till we have a good GUI runner project.

Haven't tested the new non-experimental testcentric runner (looks promising) but I suspect this belongs there as a feature request?

sidshetye:
~~Haven't tested the new non-experimental testcentric runner (looks promising) but I suspect this belongs there as a feature request?~~

Update: Just did a quick and dirty test using the non-experimental testcentric GUI runner and it already seems to support debugging when F5 launched from VS2017 (v15.7.1). Doesn't work for `netstandard2.0` but have opened [an issue](https://github.com/TestCentric/testcentric-gui/issues/44) for that. For now we'll just wait to see which GUI the nunit team decides to back next month.

CharliePoole:
Now that I moved the experimental gui into the testcentric-gui project and restructured things, this PR is orphaned. I'll look it over and figure out what to do in the next few weeks.

CharliePoole:
Closing this PR, which has been replaced by TestCentric/testcentric-gui#250

PR #186, opened 2017-05-20
Allow changing ProcessModel and DomainUsage

### LINKED ISSUE #154: Change default to run in a separate process, allowing 32-bit processes to work among other things
To reproduce, just "open" a 32-bit test assembly. An unhandled exception occurs indicating you need a 32-bit process; but there is no obvious way to do so.

Also, if this is done and the app data settings file gets saved, the crash happens on startup every time. 

Please add support for 32-bit test assemblies.

CharliePoole:
Fixes #154 

I'm pretty sure this isn't the last word on the subject, especially WRT the UI. However, it works and is ready for code review and merging so that people can use the various load settings. I still have to hook up the X86 option, but I thought it should start getting reviewed.

After discussing putting it elsewhere, I ended up leaving all three menu items on the File menu. The Process menu is invisible until after a test is loaded. Right now none of the three (Selected Runtime, ProcessModel and DomainUsage actually cause a reload. When you change one you get asked if you want to reload now. You can say no and go back to change another one. However, if it looks like that will happen often, we might want to put all of these settings into a dialog. Main thing for now is that they actually should work!

Also did some cleanup of the UiKit assembly, removing some of the exposure of internal windows forms handles. It's not all gone yet, however.

### DIFF
--- a/src/nunit-gui/Presenters/MainPresenter.cs
+++ b/src/nunit-gui/Presenters/MainPresenter.cs
@@ -74,6 +74,7 @@ namespace NUnit.Gui.Presenters
             _view.SelectedRuntime.SelectionChanged += SelectedRuntime_SelectionChanged;
             _view.ProcessModel.SelectionChanged += ProcessModel_SelectionChanged;
             _view.DomainUsage.SelectionChanged += DomainUsage_SelectionChanged;
+            _view.RunAsX86.CheckedChanged += LoadAsX86_CheckedChanged;
             _view.ExitCommand.Execute += () => Application.Exit();
 
             _view.SettingsCommand.Execute += OpenSettingsDialogCommand_Execute;
@@ -209,9 +210,18 @@ namespace NUnit.Gui.Presenters
             ChangePackageSetting(EnginePackageSettings.DomainUsage, _view.DomainUsage.SelectedItem);
         }
 
-        private void ChangePackageSetting(string key, string setting)
+        private void LoadAsX86_CheckedChanged()
         {
-            if (setting == "DEFAULT")
+            var key = EnginePackageSettings.RunAsX86;
+            if (_view.RunAsX86.Checked)
+                ChangePackageSetting(key, true);
+            else
+                ChangePackageSetting(key, null);
+        }
+
+        private void ChangePackageSetting(string key, object setting)
+        {
+            if (setting == null || setting is string && (string)setting == "DEFAULT")
                 _model.PackageSettings.Remove(key);
             else
                 _model.PackageSettings[key] = setting;
--- a/src/nunit-gui/Presenters/MainPresenter.cs
+++ b/src/nunit-gui/Presenters/MainPresenter.cs
@@ -69,7 +69,7 @@ namespace NUnit.Gui.Presenters
             _view.CloseCommand.Execute += _model.UnloadTests;
             _view.SaveCommand.Execute += _model.SaveProject;
             _view.SaveAsCommand.Execute += _model.SaveProject;
-            _view.ReloadTestsCommand.Execute += () => _model.ReloadTests();
+            _view.ReloadTestsCommand.Execute += _model.ReloadTests;
             _view.RecentProjectsMenu.Popup += RecentProjectsMenu_Popup;
             _view.SelectedRuntime.SelectionChanged += SelectedRuntime_SelectionChanged;
             _view.ProcessModel.SelectionChanged += ProcessModel_SelectionChanged;
@@ -128,9 +128,7 @@ namespace NUnit.Gui.Presenters
                     // Don't use Full suffix, but keep Client if present
                     if (text.EndsWith(" - Full"))
                         text = text.Substring(0, text.Length - 7);
-                    var menuItem = new ToolStripMenuItem(text);
-                    menuItem.Tag = runtime.ToString();
-                    dropDownItems.Add(menuItem);
+                    dropDownItems.Add(new ToolStripMenuItem(text) { Tag = runtime.ToString() });
                 }
 
                 _view.SelectedRuntime.Refresh();
@@ -138,9 +136,6 @@ namespace NUnit.Gui.Presenters
 
             // Project Menu
             _view.ProjectMenu.Enabled = _view.ProjectMenu.Visible = _model.HasTests;
-            //_view.SelectRuntimeMenu.Enabled = canCloseOrSave;
-            //_view.ProcessModel.Enabled = canCloseOrSave;
-            //_view.DomainUsage.Enabled = canCloseOrSave;
         }
 
         #endregion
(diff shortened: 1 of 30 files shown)


src/nunit.uikit/MessageDisplay.cs
```diff
@@ -49,25 +49,45 @@ public MessageDisplay(string caption)
...
+        }
+
+        public DialogResult Display(string message, string caption)
+        {
+            return Display(message, caption, MessageBoxButtons.OK);
         }
 
         public DialogResult Display(string message, MessageBoxButtons buttons)
         {
+            return MessageBox.Show(message, _defaultCaption, buttons);
```
immeraufdemhund: This could just call the overload Display to match previous 2 methods. Same applies to Error Info and Ask.

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -108,12 +112,35 @@ private void InitializeMainMenu()
...
-            _view.SelectRuntimeMenu.Enabled = canCloseOrSave;
             _view.RecentProjectsMenu.Enabled = !isTestRunning;
             _view.ExitCommand.Enabled = true;
 
+            // Special handling for available runtimes. Use model to initialize
+            // HACK: This exposes the underlying ToolStripMenuItem
+            var dropDownItems = _view.SelectRuntimeMenu.DropDownItems;
+            // Null check for when we are using a mock view to test
+            // Count check to avoid initializing twice
+            if (dropDownItems != null && dropDownItems.Count == 1)
```
CharliePoole: Yeah... I know it will work, but for an int I don't much like it for clarity.

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -108,12 +112,35 @@ private void InitializeMainMenu()
...
+            {
+                foreach (var runtime in _model.AvailableRuntimes)
+                {
+                    var text = runtime.DisplayName;
+                    // Don't use Full suffix, but keep Client if present
+                    if (text.EndsWith(" - Full"))
+                        text = text.Substring(0, text.Length - 7);
+                    var menuItem = new ToolStripMenuItem(text);
+                    menuItem.Tag = runtime.ToString();
```
CharliePoole: Good one!

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -108,12 +112,35 @@ private void InitializeMainMenu()
...
+                    dropDownItems.Add(menuItem);
+                }
+
+                _view.SelectedRuntime.Refresh();
+            }
+
             // Project Menu
             _view.ProjectMenu.Enabled = _view.ProjectMenu.Visible = _model.HasTests;
+            //_view.SelectRuntimeMenu.Enabled = canCloseOrSave;
```
CharliePoole: Changed my mind several times on this one. Finally decided that these three items should always be available, even during a test run, but not take effect immediately.

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -168,23 +195,50 @@ private void MainForm_Closing(object sender, FormClosingEventArgs ea)
...
 
-        private void OnReloadTestsCommand(string runtime)
+        private void LoadAsX86_CheckedChanged()
         {
-            _model.ReloadTests(runtime);
+            var key = EnginePackageSettings.RunAsX86;
+            if (_view.RunAsX86.Checked)
+                ChangePackageSetting(key, true);
+            else
+                ChangePackageSetting(key, null);
```
CharliePoole: I prefer not to clutter the settings, so I want to remove it rather than setting it false. In many cases, the framework interprets the absence of a setting in a different way from __any__ of the accepted values, and ChangePackageSettings is intended to work with as little knowledge of the framework as possible. I kept the "magic" value "DEFAULT" from my earlier work, but it might be better to replace that with null in the calling method as well. Also, in other kinds of settings within NUnit, there is a convention that null means delete it.

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -168,23 +195,50 @@ private void MainForm_Closing(object sender, FormClosingEventArgs ea)
...
+        }
+
+        private void ChangePackageSetting(string key, object setting)
+        {
+            if (setting == null || setting is string && (string)setting == "DEFAULT")
```
CharliePoole: Good call - I didn't think of that.

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -168,23 +195,50 @@ private void MainForm_Closing(object sender, FormClosingEventArgs ea)
...
+                _model.PackageSettings.Remove(key);
+            else
+                _model.PackageSettings[key] = setting;
+
+            string message = string.Format("New {0} setting will not take effect until you reload.\r\n\r\n\t\tReload Now?", key);
```
CharliePoole: Not a fan of string interpolation. 😞 We generally do not use . var for ints, bools, strings or other basic Types with short names.

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -213,33 +267,24 @@ private void OpenExtensionsDialogCommand_Execute()
...
-            var dropDownItems = _view.SelectRuntimeMenu.ToolStripItem.DropDownItems;
-            dropDownItems.Clear();
-            foreach (var runtime in _model.AvailableRuntimes)
-            {
-                var menuItem = runtime.DisplayName;
-                // Don't use Full suffix, but keep Client if present
-                if (menuItem.EndsWith(" - Full"))
-                    menuItem = menuItem.Substring(0, menuItem.Length - 7);
-                dropDownItems.Add(menuItem, null, (s, ea) => OnReloadTestsCommand(runtime.ToString()));
+                    var menuText = string.Format("{0} {1}", ++num, entry);
```
CharliePoole: Yuck! (Personal, not a standard)

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -213,33 +267,24 @@ private void OpenExtensionsDialogCommand_Execute()
...
         private void RecentProjectsMenu_Popup()
         {
-            var dropDownItems = _view.RecentProjectsMenu.ToolStripItem.DropDownItems;
-            dropDownItems.Clear();
-            int num = 0;
-            foreach (string entry in _model.RecentFiles.Entries)
+            // HACK: This expsoses the underlying ToolStripMenuItem
+            var dropDownItems = _view.RecentProjectsMenu.DropDownItems;
+            // Test for null, in case we are running tests with a mock
+            if (dropDownItems != null)
```
CharliePoole: Good idea

src/nunit-gui/Presenters/MainPresenter.cs
```diff
@@ -213,33 +267,24 @@ private void OpenExtensionsDialogCommand_Execute()
...
             {
-                var menuText = string.Format("{0} {1}", ++num, entry);
-                var menuItem = new ToolStripMenuItem(menuText);
-                menuItem.Click += (sender, ea) =>
+                dropDownItems.Clear();
+                int num = 0;
```
CharliePoole: Benefiting us how? 😄 See comment above.

CharliePoole:
Thanks for the review - I took most of your advice!

PR #187, opened 2017-05-23
Back out text progress notices from ProgressBar

CharliePoole:
See comment on issue #75 as to why this is being backed out.

@immeraufdemhund In addtion to my comments on the issue, I noticed in looking at the code that you have the view referencing the model, which breaks the WPF pattern, especially if we want to substitute different views at any time. This is an easy mistake to make if the components are all in the same assembly, so maybe we need some separation.

### DIFF
--- a/src/nunit-gui/Presenters/ProgressBarPresenter.cs
+++ b/src/nunit-gui/Presenters/ProgressBarPresenter.cs
@@ -85,10 +85,10 @@ namespace NUnit.Gui.Presenters
                 }
         }
 
-        private void UpdateProgress(ResultNode result)
+        private void UpdateProgress(TestNode result)
         {
             if (!result.IsSuite)
-                _progressBar.AddStatus(result.Outcome.Status);
+                _progressBar.Progress++;
         }
     }
 }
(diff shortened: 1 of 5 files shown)


immeraufdemhund:
I mean this in the most curious way possible....What do you mean? I'm visualizing the MVVM pattern (granted I self learned the pattern and don't do hardly any WPF programming anymore, hence why I'm curious) and reviewed the changes and don't see it. Second is I didn't think that the MVP pattern and MVVM pattern were that similar. If I didn't have this massive head ache right now (I'm blaming the nerf gun to the eye and not the 12 hour day) I'd explain what I think it is but I don't think it would be very coherant and even if you did reply I probably wouldn't comprehend it till tomorrow. :(

as far as reviewing this, if you wanna back it out it's all good by me.

CharliePoole:
The two are similar in the sense that they both derive from the original MVC pattern, but they have important differences. We use the passive view variation of MVP scribed here: https://martinfowler.com/eaaDev/PassiveScreen.html The presenter has references to both the view and the model and the latter two do not know about one another at all.

In the MVVM pattern, on the other hand, the View refrerences the ViewModel, which in turn references the model. 

There is a whole discussion to be had about whether we want to keep this pattern, but I think we should stick to it until we decide to do something else and not make changes too loosely. I plan to write up an issue about the arhitecture "when I have the time." 😄

PR #190, opened 2017-05-24
Separate assembly for view, allowing for multiple views

### LINKED ISSUE #189: Move View to a separate Assembly
Our intention is for the view to be decoupled from the presenter and model, which it doesn't reference. This will be easier to enforce if the view is in a separate assembly. It also will help if we want to have multiple views in future, using different underlying implementations.

The current View namespace should be combined with the UiKit assembly. The new assembly name should indicate that this is the SWF implementation of the NUnit Gui Views.

CharliePoole:
This will fix #189 when complete

** Do Not Merge - But Please Review as Commits Appear**

I think this type of review is a good idea for a change that affects the entire shape of the
application. I'll add a new comment for each commit.


CharliePoole:
For the first commit, I moved the model to a separate assembly named `nunit.gui.dll`. I also did a small amount of cleanup by removing unnecessary using statements.

This has already proven to be a useful exercise, because it revealed a couple of places where the specific gui implementation impinges on the model...

1. Some model settings use Types from System.Drawing: Point, Size and Font. Options to fix...
   * Use common types like int and string instead
   * Use separate settings for each implementation

2. In a few places we use System.MessageBox to display errors in the model
   * We should be able to do this through the view

3. The model includes command-line options, which logically belong to each exe.
   * Translate the options immediately into package settings and store them that way.

Once these items are fixed, I can remove the references to System.Drawing and System.Windows.Forms. After that, I'll move on to the more complicated job of splitting off the presenters.

immeraufdemhund:
i'm not sure if i should put my thoughts here or in #189 my eyes have been trained by work to see `some.assembly.[core|shared|interfaces|etc...]` so i'm having a hard time swallowing `nunit.gui` and how it differs from `nunit-gui` I can tell after reading the file names, That's my first initial thoughts on it. I'll keep digging around but so far I think I like the folder structure of the project. When we get more projects in the solution I'd personally like to get some solution folders.

CharliePoole:
I was suggesting we move all our comments on this issue here, because I'd like comments as it proceeds and that's easier to do in the PR. It's not how we usually do it, so any confusion is. understandable.

I agree that the naming is not good. Logically, the exe would be something like nunit-swf.exe rather than nunit-gui, but users are not accustomed to that. The old nunit gui was just nunit.exe and we could consider going back to that. AFAIC the naming is a detail we can change at any point prior to merging.

RIght now, the changes are minor - it's all just moving stuff. As I deal with the problems that surfaced, I'll have to put in more code changes that will need review. At least it's building. :-)

CharliePoole:
New commits rename `nunit.gui` to `nunit.gui.core` and move all view interfaces to the new assembly.

I was able to move (locally) only about half of the presenters without creating problems with back references from `nunit.gui.core.dll` to `nunit-gui.exe`, which we don't want, so I limited this commit to just the view interfaces.

I think the next problem to solve is that the interfaces themselves include some Windows Forms Types. I designed the app using MVP for testability, but not with an eye toward plugging in diverse views. If we want to do that, the SWF references have to go. But even if we don't go for reuse, I think it would be a cleaner interface without the windows stuff leaking out. I'll add a longer comment about this on the Epic.

immeraufdemhund:
I agree with getting rid of the SWF things from the presenters. I've been itching to add in some WPF and the other day when I had that in mind I was working towards removing the dependency on SWF. I've only been looking at your latest commit since you've pushed it for about 5 minutes now.

CharliePoole:
I'm going to create a list here of where we need to work to remove SWF dependencies. With a moderate amount of care, we can both work in this branch if you like, picking individual sub-issues to deal with.

immeraufdemhund:
yeah.. so for instance... in `IMainView`
```
        Point Location { get; set; }
        Size Size { get; set; }
        FormWindowState WindowState { get; set; }
        Font Font { get; set; }
```

I think that those sorts of things should be moved to the SWF assembly itself to save those sort of settings and not be shared in IMainView.... infact the only thing I can see (without having to dig super deep) is in IXmlView. Everytyhing else looks like it can move to its implementation.

immeraufdemhund:
Sure, that would be fine. In the end we (by we I mean I don't mind if I'm the one) to squish and rebase/cherry-pick the commits into master. I'm not able to do any work for about 6 more hours though.

CharliePoole:
Here are the places we need to remove dependency on SWF

nunit.gui.core assembly:

- [ ] GuiSettings uses `Point`, `Size` and `Font` leading IMainView and MainPresenter to need them as well.
- [ ] GuiSettings uses `Form.DefaultFont`
- [ ] The TestModel uses `MessageBox` directly, but only for not-yet-implemented messages.
- [ ] IMainView uses three windows events, passing them on to the presenter - should be our own events.
- [ ] IXmlView uses `IControlElement<RichTextBox>`

nunit.uikit assembly

- [ ] ISettingsDialog returns `DialogResult`
- [ ] IMessageDisplay exposes both `DialogResult` and `MessageBoxButtons`
- [ ] IViewElement uses `MethodInvoker`
- [ ] ITreeViewElement uses `TreeView` and `TreeNode`
- [ ] IMenu exposes `ToolStripItemCollection`
- [ ] IControlElement exposes `Point`, `Size` and `Control` itself!
- [ ] IContextMenuElement uses `ContextMenuStrip` and `ToolStripMenuItem`

@immeraufdemhund I'm going to start with the uikit stuff, so if you want to work in nunit.gui.core that will avoid interference to some extent. If you do any, watch out for use of items from uikit that may affect you and leave a comment. Let's check off items when they are committed so we know what is done and what isn't.

CharliePoole:
Created our own DialogResult and MessageBoxButtons equivalents. I selected options that can be used in all three gui platforms we have discussed.

CharliePoole:
@immeraufdemhund I see you checked off three items, but I don't see any commits. Let's not get too far out of synch!

immeraufdemhund:
yeah I just looked at where you were, and it helps me some. I'm in a different branch which was off of yours. I'll probably rebase, and squish, then switch my branch name to the same. i cleverly added a 'i' to my branch name (issue-189i)

immeraufdemhund:
oh, I can't figure out how to make the settings persist for my new struct. I'm going to make a work in progress commit which is where i'm attacking the `GuiSetttings uses Point, Size, and Font`

I'm also not sure how i'm going to attack font. I'm thinking another struct with a string and float for name and size...

immeraufdemhund:
I see that in your commit you kept the implementation for the MessageDialog in UiKit. Did we want to move that to `nunit-gui`?  And as a related side note. If we build up a large list of interfaces, can we use the engine to do the constructor injection if we tell it what implementation is for which interface? (Guessing that is what `ServiceLocator` is for)

CharliePoole:
@immeraufdemhund Well, my idea was that we work in the same branch, committing frequently and reviewing one another's commits. I can do the same in your branch instead, provided you tell me when you commit just as I am doing.

CharliePoole:
I see our notes crossed and your commits are in the PR now. I'll review in the morning.

CharliePoole:
I didn't do anything different with NuNit yet because we were going to work very narrowly at removing the SWF refs first. I want to stick to plan, i

CharliePoole:
I would do don't as a string and point and size as a type. See format in your existing settings file.

Back to bed now!

src/nunit-gui/Program.cs
```diff
@@ -79,7 +80,8 @@ static void Main(string[] args)
             if (options.InternalTraceLevel != null)
                 testEngine.InternalTraceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), options.InternalTraceLevel);
 
-            var model = new TestModel(testEngine, options);
+            IMessageDisplay messageDisplay = new MessageDisplay();
+            var model = new TestModel(testEngine, options, messageDisplay);
```
CharliePoole: I'm afraid I may not have stated the problem clearly enough. It's not that the model should use a platform-independent method to issue messages, it's that it shouldn't be issueing messages at all. Messages are issued by the view, when called by the presenter. The model should not know anything about the view.

Additionally, in the way you implemented it, you didn't really remove the windows dependency because MessageDisplay is part of the SWF implementation.

I'm going to remove this commit, but I wanted to leave you the comments first.

src/nunit.gui.core/Model/TestModel.cs
```diff
@@ -202,12 +205,12 @@ public void NewProject()
 
         public void NewProject(string filename)
         {
-            MessageBox.Show("It's not yet decided if we will implement saving of projects. The alternative is to require use of the project editor, which already supports this.", "Not Yet Implemented");
+            _messageDisplay.Info("It's not yet decided if we will implement saving of projects. The alternative is to require use of the project editor, which already supports this.", "Not Yet Implemented");
```
CharliePoole: Rather than issuing a message here, we should call _view.MessageDisplay in the presenter where we call this.

src/nunit.gui.core/Views/IMainView.cs
```diff
@@ -36,8 +47,8 @@ public interface IMainView : IView
         FormWindowState WindowState { get; set; }
         Font Font { get; set; }
 
-        event FormClosingEventHandler FormClosing;
-        event DragEventHandler DragDrop;
+        event MainViewClosingEvent MainViewClosing;
```
CharliePoole: This can be `event CommandHandler Closing` 

`CommandHandler` is our view event handler with no args.`
`Closing` is sufficient as the name since you can only refer to it using an instance of the main view.

src/nunit.gui.core/Views/IMainView.cs
```diff
@@ -36,8 +47,8 @@ public interface IMainView : IView
...
+        event FilesDragAndDroppedEvent FilesDragAndDropped;
```
CharliePoole: The handler is a handler, not an event. 😄 I would keep the same names as windows, but in our namespace.

src/nunit.gui.core/Views/IXmlView.cs
```diff
@@ -33,7 +31,7 @@ public interface IXmlView : IView
         bool Visible { get; set; }
         string Header { get; set; }
         IViewElement XmlPanel { get; }
-        IControlElement<RichTextBox> XmlTextBox { get; }
+        IXmlTextBoxView XmlTextBox { get; }
```
CharliePoole: This solves the problem. I'll need to come back and integrate it with the the controls in UiKit.Elements in a subsequent commit.

build.cake
```diff
@@ -116,87 +116,87 @@ Task("Build")
...
+            BIN_DIR + "nunit.uikit.dll",
+            BIN_DIR + "nunit.uikit.pdb",
+            BIN_DIR + "nunit.engine.api.dll",
+            BIN_DIR + "nunit.engine.dll",
+            BIN_DIR + "nunit.engine.addins",
+            BIN_DIR + "Mono.Cecil.dll"
+        };
+
+        Zip(BIN_DIR, File(ZIP_PACKAGE), zipFiles);
+    });
```
CharliePoole: In general, don't mix changes like this in with substantive changes to other files, especially when working cooperatively. We can stand this one, but see other comments.

src/nunit-gui/Views/MainForm.Designer.cs
```diff
@@ -907,21 +898,21 @@ private void MainForm_DragEnter(object sender, DragEventArgs e)
...
+        private System.Windows.Forms.ToolStripMenuItem inProcessToolStripMenuItem;
+        private System.Windows.Forms.ToolStripMenuItem singleProcessToolStripMenuItem;
+        private System.Windows.Forms.ToolStripMenuItem multipleProcessToolStripMenuItem;
+        private System.Windows.Forms.ToolStripSeparator toolStripSeparator2;
+        private System.Windows.Forms.ToolStripMenuItem loadAsX86ToolStripMenuItem;
+        private System.Windows.Forms.ToolStripMenuItem domainUsageToolStripMenuItem1;
+        private System.Windows.Forms.ToolStripMenuItem defaultDomainToolStripMenuItem;
+        private System.Windows.Forms.ToolStripMenuItem singleDomainToolStripMenuItem;
+        private System.Windows.Forms.ToolStripMenuItem multipleDomainToolStripMenuItem;
+        private System.Windows.Forms.ToolStripMenuItem notYetImplementedToolStripMenuItem;
```
CharliePoole: Why these changes?

src/nunit-gui/Views/MainForm.cs
```diff
@@ -163,5 +163,11 @@ protected override void OnDragDrop(DragEventArgs drgevent)
             if (files != null)
                 FilesDragAndDropped?.Invoke(files);
         }
+
+        protected override void OnDragEnter(DragEventArgs drgevent)
+        {
+            if (drgevent.Data.GetDataPresent(DataFormats.FileDrop))
+                drgevent.Effect = DragDropEffects.Copy;
+        }
```
CharliePoole: This should be part of drag and drop I guess.

I'm seeing three different things in this commit. I'd like to leave out one. I can do it but it's more work than if you kept things isolated.

CharliePoole:
@immeraufdemhund In case you are still online... I'm about to download your changes and remove one commit plus part of another, as described in my comments. You should probably stash any local changes you have not yet committed before pulling. I'll post again when done.

CharliePoole:
@immeraufdemhund It's all updated now with a forced push. Not really that great but seems like the best way to handle it at this point. You'll have to be careful with any local changes when you next pull.

src/nunit-gui/Views/MainForm.Designer.cs
immeraufdemhund: in the commit that was added a while ago that added support for Drag and drop the user added code directly to the auto-generated designer file. When they did they also added a using to the top of the file. Later in one of your commits you change that form and so it was generated again. It dropped the event (which is why it stopped working), so this was a move back to "default" generated code. Winforms usually has no usings in the designer file so I took out the using and added back in the fully qualified name.

src/nunit-gui/Views/MainForm.cs
immeraufdemhund: sorry I have a habbit of fixing things that I know are wrong and are easy to fix instead of leaving a note. Really this and the designer file can be moved to a different PR for a different day in the long run. Boy scout rule was kicking in on these.

immeraufdemhund:
>@immeraufdemhund It's all updated now with a forced push. Not really that great but seems like the best way to handle it at this point. You'll have to be careful with any local changes when you next pull.

At work for the past 2 months 4 of us have been working on the same project. this was nothing in comparison. I've become accustomed to fetching and looking at gitk before doing any sorta merging or pulling. :smile_cat:  Thank you though for the heads up. As a side note, if you wanted,  you or I  could cherry-pick the  commit with cake and drag and drop onto master and get them out of this pull request. As you noted they don't fit in this pull request or issue.

src/nunit.gui.core/Model/TestModel.cs
immeraufdemhund: yeah sorry about that.. I'll have to draw up a diagram and think about what i'm doing when i'm playing with the models and views and presenters. I've still not got the pattern fully committed to memory.

immeraufdemhund:
Hopefully these last two commits are what you are seeking. I tried something new with the TypeConverter, but it plays in with how SettingsService gets/sets values. I've tested it locally and it appears to be working correctly (including font.. that is fun changing the default Font through the settings file).

immeraufdemhund:
I kept getting the feeling last night that a few of the classes I made should have been in the UIKit. Did you want me to have a separate commit, or did you want to move the classes over?

src/nunit.gui.core/Model/Settings/GuiSettings.cs
```diff
@@ -59,18 +59,11 @@ public class MainFormSettings : SettingsWrapper
     {
         public MainFormSettings(ISettings settingsService) : base(settingsService, "Gui.MainForm") { }
 
-        private const string locationKey = "Location";
-        public Point Location
+        private const string windowShapeKey = "LocationAndSize";
+        public WindowShape LocationAndSize
```
CharliePoole: To be honest, I'm not crazy about this idea. Location and Size are concepts in most widget platforms I'm familiar with. I haven't seen a WindowShape before. Is this something from WPF?

src/nunit.gui.core/Model/Settings/GuiSettings.cs
```diff
@@ -81,9 +74,9 @@ public bool Maximized
         }
 
         private const string fontKey = "Font";
-        public Font Font
+        public ViewFont Font
```
CharliePoole: Since we can only store a string, why not just leave it string and have either the View or the platform-specific part of the Presenter translate it? Someone has to translate it anyway.

src/nunit.gui.core/Model/Settings/GuiSettings.cs
immeraufdemhund: no, i just figured it was worth combining. I forget what it it is off the top of my head in WPF, but I think the concept is the same as SWF

CharliePoole:
@immeraufdemhund See my comments on Location, Size and Font changes.

I wouldn't worry much about what goes in UiKit, since I'm planning to eliminate it.

This PR is getting pretty big and I'd like to merge some of the changes and rebase. Before we can do that, somebody has to review my code, or at least I would like them to! I'd like to start with the first five commit so we need code review on those and - even more important - agreement that this restructuring is the direction we want to go. The first five commits create and substantially populate nunit.core.gui.

If it's too hard to just review commits, I can create a new PR that incorporates only those first five. In addition to @immeraufdemhund reviewing, I'd appreciate it if @jnm2 could take a look.

Guys, let me know if you prefer a separate PR for the review.

src/nunit.gui.core/Model/Settings/GuiSettings.cs
immeraufdemhund: ? I'm not sure what you mean by we can only store it as a string. With the TypeConverter added it gets saved and restored from this

```
<?xml version="1.0"?>
<NUnitSettings>
  <Settings>
    <Setting name="Gui.MainForm.Font" value="Microsoft Sans Serif;8.25" />
    <Setting name="Gui.MainForm.Location" value="10, 10" />
    <Setting name="Gui.MainForm.LocationAndSize" value="0, 0, 545, 454" />
    <Setting name="Gui.MainForm.Maximized" value="False" />
    <Setting name="Gui.MainForm.Size" value="1349, 613" />
    <Setting name="Gui.RecentProjects.File1" value="C:\Git\survey-design\mass-update\MassUpdate.AcceptanceTests\bin\Debug\MassUpdate.AcceptanceTests.dll" />
  </Settings>
</NUnitSettings>
```
and I can manually change it to say consolas size 14 and get a massivly large screen 

image

CharliePoole:
On second thought... I will have to create a separate branch to merge anyway. So watch for a new PR shortly!

immeraufdemhund:
hey on that note, lets do yet another branch and pull request for the cake and drag and drop fix. just to help keep things separate

CharliePoole:
My plan was to do just that, merging in everything we can and then rebasing this PR to show only what is still unmerged. I think it's easier to review a whole PR than just a few commits.

Please review PR #193

immeraufdemhund:
No I haven't looked at it. I saw the diverged commit and thought about getting into. Than I started looking into wcf and what not. I'll take a gander tonight

CharliePoole:
Most of this work has been completed now in separate PRs and the only remaining item (replacing RichEditTextBox) is out of date. Closing this PR.

PR #193, opened 2017-05-27
Separate model and view interfaces into separate assembly

CharliePoole:
This PR contains the first ~~five~~four commits out of PR #190. It creates a new assembly target, `nunit.gui.core`, intended to hold those parts of the application that will not vary based on the gui toolkit we use.

Further splits are possible as we go along.


immeraufdemhund:
sorry about the delay on this, rough day yesterday. i'm starting to review the commits now.

immeraufdemhund:
I personally would like to see e5c418a and c86b478 squished (the first two commits).

CharliePoole:
@jnm2 I was particularly looking for your review of the entire concept of separating the platform dependent and independent parts of the gui into separate assemblies. 

Lately, I'm thinking we should have even more separation, putting the model into it's own assembly. If the model assembly has no reference to the presenter or view, then it's impossible to make the mistake of using something that breaks the design pattern. Of course, the alternative is that we be more disciplined about it. 😄 

FYI, second commit was a result of code review in the original PR.

To give us time to make this decision, I plan to hold it back from the 0.4 release, which I expect to do in a day or two.

immeraufdemhund:
That would help me!



Sent from my Verizon, Samsung Galaxy smartphone


-------- Original message --------

jnm2:
Erm... finally have time to sit down and think about this. Actually I may have a lot of time to sit down. I'll find out tomorrow if I've torn a ligament in my knee.

jnm2:
> If the model assembly has no reference to the presenter or view, then it's impossible to make the mistake of using something that breaks the design pattern. Of course, the alternative is that we be more disciplined about it. 

For me I'd prefer an analyzer to a separate assembly. I'm usually in favor of augmenting discipline with hard structure, but I dislike separating things into layers. Whether using projects or even Views/Models/Presenters folders. I haven't seen the new version of Jimmy Bogard's talk [Slices not Layers](https://www.youtube.com/watch?v=wTd-VcJCs_M) but it's probably as good as the first one. When I do WinForms MVVM (which is most of the time) I have each view .cs file immediately next to each view model .cs file. I'm still moving further this direction as time goes on.

You're obviously doing things a little differently. For example, I got away from the interface per view as fast as I could. I always had exactly one view. The coupling between the view and the interface was brittle and high maintenance. It felt like it was always getting in the way of development. But in your case, you intend multiple identically-behaving implementations of each view so it makes sense to enforce that through interfaces.

CharliePoole:
Both this pull and #190 have gotten really stale, with a number of conflicting changes added to master.

I redid the first four commits from #190 as two commits here. All it does is move things around, creating a new nunit.gui.core assembly that holds both the model and the interfaces for the views. I'd like to get this merged before much more happens. We'll keep #190 around for a while in order to cherry-pick changes from it and then close it without merging.

src/nunit.gui.core/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
immeraufdemhund: Change AssemblyCompany and AssemblyCopyright.

src/nunit.gui.core/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("nunit.gui")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Hewlett-Packard")]
+[assembly: AssemblyProduct("nunit.gui")]
```
immeraufdemhund: Consider changing to match assembly

src/nunit.gui.core/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("nunit.gui")]
```
immeraufdemhund: Consider changing to match assembly

src/nunit.gui.core/Properties/AssemblyInfo.cs
immeraufdemhund: I am so bad with where to "add" a comment. Sorry

CharliePoole:
Fixed assemblyinfo by adding a common assemblyinfo file for all six assemblies.

PR #194, opened 2017-05-28
Add agent files to package

### LINKED ISSUE #192: Nuget package for engine doesn't have agent config files
The config does some pretty critical stuff and this absence could well explain a number of ongoing issues we have when running under the agent.

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <!--
   Nunit-agent only runs under .NET 2.0 or higher. 
   The setting    useLegacyV2RuntimeActivationPolicy only applies 
   under .NET 4.0 and permits use of mixed mode assemblies, 
   which would otherwise not load correctly. 
  -->
  <startup useLegacyV2RuntimeActivationPolicy="true">
    <!--
     Nunit-agent is normally run by the console or gui
     runners and not independently. In normal usage, 
     the runner specifies which runtime should be used.
     
     Do NOT add any supportedRuntime elements here, 
     since they may prevent the runner from controlling 
     the runtime that is used!
    -->
  </startup>

  <runtime>
    <!-- Ensure that test exceptions don't crash NUnit -->
    <legacyUnhandledExceptionPolicy enabled="1" />

    <!-- Run partial trust V2 assemblies in full trust under .NET 4.0 -->
    <loadFromRemoteSources enabled="true" />

  </runtime>
  
</configuration>
```

I'm creating this issue in both the gui and the engine projects. The gui needs a temporary fix on it's special feed. The engine needs a permanent fix for everyone else.

CharliePoole:
Fixes #192

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -173,7 +173,11 @@ Task("PackageZip")
 			BIN_DIR + "nunit.engine.api.dll",
 			BIN_DIR + "nunit.engine.dll",
 			BIN_DIR + "nunit.engine.addins",
-			BIN_DIR + "Mono.Cecil.dll"
+			BIN_DIR + "Mono.Cecil.dll",
+			BIN_DIR + "nunit-agent.exe",
+			BIN_DIR + "nunit-agent.exe.config",
+			BIN_DIR + "nunit-agent-x86.exe",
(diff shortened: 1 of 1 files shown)


mikkelbu:
LGTM

PR #195, opened 2017-05-28
Use ShadowCopy to load tests

### LINKED ISSUE #170: Test assembly locked
Would maybe be resolved by #43 (reload on change), but currently the GUI locks the Test Assembly and so cannot be overwritten during the build without closing the application.

CharliePoole:
Fixes #170

### DIFF
--- a/src/nunit-gui/Model/TestModel.cs
+++ b/src/nunit-gui/Model/TestModel.cs
@@ -368,7 +368,7 @@ namespace NUnit.Gui.Model
                 case "test-suite":
                     result = new ResultNode(xmlNode);
                     _resultIndex[result.Id] = result;
-                    SuiteFinished?.Invoke(new TestResultEventArgs(TestAction.TestFinished, result));
+                    SuiteFinished?.Invoke(new TestResultEventArgs(TestAction.SuiteFinished, result));
                     break;
 
                 case "test-run":
--- a/src/nunit-gui/Model/TestModel.cs
+++ b/src/nunit-gui/Model/TestModel.cs
@@ -240,8 +240,7 @@ namespace NUnit.Gui.Model
             _files = null;
             _resultIndex.Clear();
 
-            if (TestUnloaded != null)
-                TestUnloaded(new TestEventArgs(TestAction.TestUnloaded));
+            TestUnloaded(new TestEventArgs(TestAction.TestUnloaded));
         }
 
         public void ReloadTests()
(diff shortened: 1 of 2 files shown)


src/nunit-gui/Model/TestModel.cs
```diff
@@ -240,8 +240,7 @@ public void UnloadTests()
             _files = null;
             _resultIndex.Clear();
 
-            if (TestUnloaded != null)
-                TestUnloaded(new TestEventArgs(TestAction.TestUnloaded));
```
mikkelbu: Why the removal of the guard? (I guess that right now we can be sure that there are handlers associated with the event, but that might not be case in the future?)

src/nunit-gui/Views/ProgressBarView.Designer.cs
```diff
@@ -55,6 +55,5 @@ private void InitializeComponent()
         #endregion
 
         private NUnit.UiKit.Controls.NUnitProgressBar testProgressBar;
-        private System.Windows.Forms.Label testStatusesLabel;
```
mikkelbu: Seems unrelated, but I'm fine with "cleaning up" :)

mikkelbu:
Looks correct. I only have some minor questions, but please merge :)

src/nunit-gui/Views/ProgressBarView.Designer.cs
CharliePoole: Warning gives an error on travis, so it has to be cleaned.

src/nunit-gui/Model/TestModel.cs
CharliePoole: Actually, I intended to use ? operator. Good catch. I'll check other events as well.

src/nunit-gui/Model/TestModel.cs
```diff
@@ -350,39 +348,33 @@ public void OnTestEvent(string report)
...
-                        TestFinished(new TestResultEventArgs(TestAction.TestFinished, result));
+                    TestFinished?.Invoke(new TestResultEventArgs(TestAction.TestFinished, result));
                     break;
 
                 case "test-suite":
                     result = new ResultNode(xmlNode);
                     _resultIndex[result.Id] = result;
-                    if (SuiteFinished != null)
-                        SuiteFinished(new TestResultEventArgs(TestAction.TestFinished, result));
+                    SuiteFinished?.Invoke(new TestResultEventArgs(TestAction.TestFinished, result));
```
mikkelbu: All the translations to conditional member access looks correct (and also we are now threadsafe ;) ). The only problem I discovered was that we are using the wrong `TestAction`. It should be `TestAction.SuiteFinished` not `TestAction.TestFinished`.

CharliePoole: Dang!

All the translations are correct, 'cause I let VS do it this time. 😄

PR #197, opened 2017-05-30
Multiple assert support

### LINKED ISSUE #140: Add support for Multiple Asserts
NUnit 3.6 will support multiple assertions in a test. We should display all of the assert results under the test case to which they apply.

CharliePoole:
Fixes #140

I combined the two fields Message and StackTrace on the TestPropertiesView and I'm displaying all the results found. The display could use some color or at least bold fonts but I don't want to do it until we resolve whether we are keeping the ExpandingLabel control, because it will be done differently depending on whether we keep or drop it. I did improve the TipWindow display a little bit.

For testing this, if you want to see it work. the NUnit 3 adapter demo program is a good choice as it has a number of tests with multiple failing asserts, failures plus warnings and failures plus errors.

### DIFF
--- a/src/nunit-gui/Presenters/TestPropertiesPresenter.cs
+++ b/src/nunit-gui/Presenters/TestPropertiesPresenter.cs
@@ -166,11 +166,36 @@ namespace NUnit.Gui.Presenters
                 sb.AppendFormat("{0}) {1}\n", ++index, assertion.Status);
                 sb.AppendLine(assertion.Message);
                 if (assertion.StackTrace != null)
-                    sb.AppendLine(assertion.StackTrace);
+                    sb.AppendLine(AdjustStackTrace(assertion.StackTrace));
 
             }
 
             _view.Assertions = sb.ToString();
+        }
+
+        // Some versions of the framework return the stacktrace
+        // without leading spaces, so we add them if needed.
+        // TODO: Make sure this is valid across various cultures.
+        private const string LEADING_SPACES = "   ";
+
+        private static string AdjustStackTrace(string stackTrace)
+        {
+            // Check if no adjustment needed. We assume that all
+            // lines start the same - either with or without spaces.
+            if (stackTrace.StartsWith(LEADING_SPACES))
+                return stackTrace;
+
+            var sr = new StringReader(stackTrace);
+            var sb = new StringBuilder();
+            string line = sr.ReadLine();
+            while (line != null)
+            {
+                sb.Append(LEADING_SPACES);
+                sb.AppendLine(line);
+                line = sr.ReadLine();
+            }
+
+            return sb.ToString();
         }
 
         private void DisplayOutput(ResultNode resultNode)
--- a/src/nunit-gui/Presenters/TestPropertiesPresenter.cs
+++ b/src/nunit-gui/Presenters/TestPropertiesPresenter.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Text;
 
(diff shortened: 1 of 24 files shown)


src/nunit.uikit/Controls/TipWindow.cs
```diff
@@ -177,6 +177,13 @@ protected override void OnLoad(System.EventArgs e)
 
 			Size sizeNeeded = Size.Ceiling( g.MeasureString( tipText, Font, layoutArea ) );
 
+            // If the needed width is smaller than that of the original label,
```
immeraufdemhund: This whole file has tabs, you used spaces. not super important. For the purpose of consistency can this switch to tabs? :cry:

src/nunit-gui/Presenters/TestPropertiesPresenter.cs
```diff
@@ -185,5 +202,29 @@ private string TrimMessage(string message)
         }
 
         #endregion
+
+        #region Nested AssertionResult Class
```
immeraufdemhund: I feel like this file is getting too large anyway, and now with a region added with a class in it just kind of rubs me the wrong way. I'm assuming it's here with the thought of when this all gets separated it will be easier... but it still sort of bugs me.

src/nunit-gui/Presenters/TestPropertiesPresenter.cs
```diff
@@ -105,27 +106,43 @@ private void DisplayTestProperties()
                     _view.ElapsedTime = resultNode.Duration.ToString("f3");
                     _view.AssertCount = resultNode.AssertCount.ToString();
 
-                    var assertions = resultNode.Xml.SelectNodes("assertions/assertion");
-                    XmlNode message = null;
-                    XmlNode stackTrace = null;
+                    var assertionNodes = resultNode.Xml.SelectNodes("assertions/assertion");
```
immeraufdemhund: The nesting depth and changes I feel would be better understood with a few more private methods to help break up this complex area

src/nunit.uikit/Controls/TipWindow.cs
immeraufdemhund: ugh, i should have chose the bottom of the change instead of the top.. then this woudl probably make more sense. I wish that GitHub allowed for multi-line comments

immeraufdemhund:
Last comment before i go to the other piece. This change mentions added multiple assert support. I would like to see a multiple assert in the mock assembly so i can pull it up here and see how this change makes it look.

CharliePoole:
I'll do some refactoring as you suggest. 

As for mock-assembly, it has traditionally been used to test counting of errors, failures, etc. and I've pushed back in the past when people added code to it to demonstrate various other features. I've been using NUnit3TestDemo.dll for the purpose of showing what multiple asserts, warnings, etc. look like. It's part of the nunit 3 adapter project, but it could easily be copied here for that use. What do you think?

CharliePoole:
Regarding TipWIndow, I'd rather not spend any more time cleaning it up until we decide whether or not to keep it.

immeraufdemhund:
>  It's part of the nunit 3 adapter project, but it could easily be copied here for that use. What do you think?

yes. I pulled down the adapter project and ran cake to build it and get the assembly and it is failing when it gets to the NUnit3TestAdapterInstall.csproj (I initially pasted the error, but I don't want to clutter up this pull request with non-related things... HEY I'M GETTING BETTER! Strange that it isn't working because it looks like it is looking for vs2017 which I have installed)

src/nunit-gui/Presenters/TestPropertiesPresenter.cs
CharliePoole: See if it's easier to understand now.

CharliePoole:
@immeraufdemhund I was a bit afraid that the idea of tip window and expanding fields on hover was a bit old-fashioned. But since you seem to like the idea, I'll clean it up a bit and keep it for now.

CharliePoole:
I think that should cover all the comments - ready to review again.

CharliePoole:
They appeared only after I added the new demo project, which makes no sense at all. I'm looking into it.

CharliePoole:
Silly mistake!

CharliePoole:
Interesting... I just noticed that Visual Studio uses a very similar display to my TipWindow when you hover over the stack trace in an exception notice.

CharliePoole:
Just a few visual changes while waiting for re-review:
* All three large text boxes are now fixed font - looks better that way.
* Top border moved up one pixel, left border left one pixel.
* Increase padding within window around the text.

PR #202, opened 2017-06-09
fixes drag and drop

immeraufdemhund:
closes issue #155

### DIFF
--- a/src/nunit-gui/Views/MainForm.cs
+++ b/src/nunit-gui/Views/MainForm.cs
@@ -144,5 +144,10 @@ namespace NUnit.Gui.Views
         {
             statusBarView.Visible = statusBarToolStripMenuItem.Checked;
         }
+
+        protected override void OnDragEnter(DragEventArgs drgevent){
+            if(drgevent.Data.GetDataPresent(DataFormats.FileDrop))
+                drgevent.Effect = DragDropEffects.Copy;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/nunit-gui/Views/MainForm.cs
```diff
@@ -144,5 +144,10 @@ private void statusBarToolStripMenuItem_CheckedChanged(object sender, System.Eve
         {
             statusBarView.Visible = statusBarToolStripMenuItem.Checked;
         }
+
+        protected override void OnDragEnter(DragEventArgs drgevent){
+            if(drgevent.Data.GetDataPresent(DataFormats.FileDrop))
+                drgevent.Effect = DragDropEffects.Copy;
+        }
```
CharliePoole: My understanding is that this is merely the visual effect to tell the user that a drop is possible. Was the actual drop already working?

immeraufdemhund: Let me double check. That was the only pending change I had and when I built and dragged and dropped it worked on my machine

immeraufdemhund: ok, so yes you are right that you need both the visual effect and the subscribing to the event to capture getting the file names that are dropped, however that subscription is in the presenter. When we eventually move the SWF stuff out of the presenter we'll have to also override the DragDrop event to fire the new event in the View.

As a partial side note because I double checked there is extra stuff in the designer that I should delete. I'll do that and push up another commit

CharliePoole:
Thanks! I've merged it so we can all see it in our builds and in 0.4. Given that this is pre-production that seems like the best way to go.

PR #203, opened 2017-06-21
Prepare for release 0.4

CharliePoole:
This PR updates the CHANGES file and changes version numbers for the release. If there is anything missing that should be included in the release, please add comment.

I expressly held back any changes to how the files are organized for after the release.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,27 @@
-﻿NUnit-Gui 0.3 - January 1, 2017
+﻿NUnit-Gui 0.4 - June 21, 2017
+
+Issues Resolved
+
+ * 14     Display relative path in GUI's tree (option)
+ * 26     Explicit tests are executed if a category is excluded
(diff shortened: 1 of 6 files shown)


CharliePoole:
In testing, I saw we have been including a spurious `.addins` file that points to extensions, which aren't present! We'll eventually put back an `.addins` file, but it will need to really point at something!

grokky1:
@CharliePoole Does this work with .NET Core?

CharliePoole:
@grokky1 Not yet. The GUI uses the NUnit engine so it will work with .NET Core when the engine does, that is, at the same time the console runner can run .NET Core tests. Right now, we are focusing primarily on UI design, but we will keep up with new engine versions as they come out.

grokky1:
@CharliePoole Sounds good.

It's a real pain to use VS' "Test Explorer" which is slow and cumbersome. Can't wait for this gui runner (and the console runner) to support Core!

PR #210, opened 2017-07-05
Create chocolatey package

### LINKED ISSUE #205: Create a Chocolatey package for the GUI
I'm viewing Chocolatey as the primary method of distributing the GUI, once it is separated.

Rationale... MSIs are a pain! NuGet packages for a gui don't make a lot of sense. We can hold onto the zip for folks who have some special need to get just the files.

The GUI can install itself under Chocolatey as a standalone package and look for extensions that have also been installed under chocolatey. We may need to create some of the extension installs as well, and contribute to the NUnit Project. This allows a way of working where we don't do any bundling - users decide what extensions they want.

CharliePoole:
Fixes #205

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -182,6 +182,38 @@ Task("PackageZip")
         Zip(BIN_DIR, File(ZIP_PACKAGE), zipFiles);
     });
 
+Task("PackageChocolatey")
+    .IsDependentOn("Build")
+    .Does(() =>
+    {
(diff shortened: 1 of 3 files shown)


CharliePoole:
A chocolatey package for 0.4 has been uploaded and is waiting for moderation. If changes are needed, I'll put them into the 0.5 build.

PR #215, opened 2017-08-16
Remove Windows references from IMessageDisplay

CharliePoole:
Continued work on issue #189. I'm doing this a bit at a time in order to get things committed quickly and avoid conflicts.


CharliePoole:
@immeraufdemhund I'm going to go ahead and merge this as I have another PR that depends on it. If you find anything to comment on, I can make corrections in the next PR.

immeraufdemhund:
looked good to me. One thing i wondered about is if our enums have the same values if we can't just cast from one enum to another instead of using switches.

CharliePoole:
The switch or equivalent is needed because of `default:` - there are more values in the Windows enum than exist in our own `DialogResult`.

PR #219, opened 2017-08-25
Move settings out of model

CharliePoole:
Continuing isolation of model... partial fix to #189

### DIFF
--- a/src/nunit.gui.core/Model/TestModel.cs
+++ b/src/nunit.gui.core/Model/TestModel.cs
@@ -196,14 +196,17 @@ namespace NUnit.Gui.Model
             NewProject("Dummy");
         }
 
+        public const string PROJECT_SAVE_MESSAGE =
+            "It's not yet decided if we will implement saving of projects. The alternative is to require use of the project editor, which already supports this.";
+
         public void NewProject(string filename)
         {
-            throw new NotImplementedException();
+            throw new NotImplementedException(PROJECT_SAVE_MESSAGE);
         }
 
         public void SaveProject()
         {
-            throw new NotImplementedException();
+            throw new NotImplementedException(PROJECT_SAVE_MESSAGE);
         }
 
         public void LoadTests(IList<string> files)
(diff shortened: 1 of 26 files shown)


src/nunit.gui.core/Model/TestModel.cs
```diff
@@ -202,12 +198,12 @@ public void NewProject()
 
         public void NewProject(string filename)
         {
-            MessageBox.Show("It's not yet decided if we will implement saving of projects. The alternative is to require use of the project editor, which already supports this.", "Not Yet Implemented");
+            throw new NotImplementedException();
```
immeraufdemhund: Why this change? I understand the gist of it, but I don't see where we are catching this type of exception to show the user. So when this is called by the MainPresenter would this cause an unhandled exception to occur. Same for `SaveProject`

src/tests/Presenters/UserSettingsFake.cs
```diff
@@ -28,6 +28,9 @@
 
 namespace NUnit.Gui.Presenters
 {
+    // TODO: With latest changes, this is no longer used.
+    // Keeping it for a while, as we may end up wanting
+    // to write more tests that use it.
     class UserSettingsFake : ISettings
```
immeraufdemhund: Maybe consider marking this as obsolete with the same message, then either delete it or flip the flag to make it cause a compile error.

src/nunit.gui.core/Model/TestModel.cs
CharliePoole: Good catch. I intended to move the message display to the presenter. This gets System.Windows.Forms out if the model.

src/tests/Presenters/UserSettingsFake.cs
CharliePoole: Since it's in our tests, I see no harm in it for now. This is part of a series of changes to the same code so I'll be coming back here soon.

PR #221, opened 2017-08-26
Eliminate direct use of engine for runtime frameworks

### LINKED ISSUE #209: Utilise IAvailableFrameworks interface over RuntimeFramework class in engine
I'm currently looking at refactoring the RuntimeFramework classes in the NUnit engine to improve thread-safety. (https://github.com/nunit/nunit-console/issues/173)

The GUI currently directly references the RuntimeFramework class, to access framework features. https://github.com/nunit/nunit-console/pull/249 adds the required functionality to the nunit.engine.api assembly, to allow the GUI to access the features from the interfaces, instead of directly.

Time permitting, I hope to refactor RuntimeFramework (thus breaking the GUI's current reliance on the RuntimeFramework class) in NUnit Engine 3.8.

CharliePoole:
Fixes #209 

The GUI no longer uses the `RuntimeFramework` type but instead accesses the `IAvailableRuntimes` service. As a result, the reference to the engine itself has been removed from both the `nunit.gui.core` and `nunit-gui` projects.

### DIFF
--- a/src/nunit.gui.core/EnginePackageSettings.cs
+++ b/src/nunit.gui.core/EnginePackageSettings.cs
@@ -1,0 +1,157 @@
+﻿// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit
+{
+    /// <summary>
+    /// EngineSettings contains constant values that
+    /// are used as keys in setting up a TestPackage. 
+    /// Values here are used in the engine, and set by the runner.
+    /// Setting values may be a string, int or bool.
+    /// </summary>
+    public static class EnginePackageSettings
+    {
+        #region Engine Settings - Used by the Engine itself
+
+        /// <summary>
+        /// The name of the config to use in loading a project.
+        /// If not specified, the first config found is used.
+        /// </summary>
+        public const string ActiveConfig = "ActiveConfig";
+
+        /// <summary>
+        /// Bool indicating whether the engine should determine the private
+        /// bin path by examining the paths to all the tests. Defaults to
+        /// true unless PrivateBinPath is specified.
+        /// </summary>
+        public const string AutoBinPath = "AutoBinPath";
+
+        /// <summary>
+        /// The ApplicationBase to use in loading the tests. If not 
+        /// specified, and each assembly has its own process, then the
+        /// location of the assembly is used. For multiple  assemblies
+        /// in a single process, the closest common root directory is used.
+        /// </summary>
+        public const string BasePath = "BasePath";
+
+        /// <summary>
+        /// Path to the config file to use in running the tests. 
+        /// </summary>
+        public const string ConfigurationFile = "ConfigurationFile";
+
+        /// <summary>
+        /// Flag (bool) indicating whether tests are being debugged.
+        /// </summary>
+        public const string DebugTests = "DebugTests";
+
+        /// <summary>
+        /// Bool flag indicating whether a debugger should be launched at agent 
+        /// startup. Used only for debugging NUnit itself.
+        /// </summary>
+        public const string DebugAgent = "DebugAgent";
+
+        /// <summary>
+        /// Indicates how to load tests across AppDomains. Values are:
+        /// "Default", "None", "Single", "Multiple". Default is "Multiple"
+        /// if more than one assembly is loaded in a process. Otherwise,
+        /// it is "Single".
+        /// </summary>
+        public const string DomainUsage = "DomainUsage";
+
+        /// <summary>
+        /// The private binpath used to locate assemblies. Directory paths
+        /// is separated by a semicolon. It's an error to specify this and
+        /// also set AutoBinPath to true.
+        /// </summary>
+        public const string PrivateBinPath = "PrivateBinPath";
+
+        /// <summary>
+        /// The maximum number of test agents permitted to run simultaneously. 
+        /// Ignored if the ProcessModel is not set or defaulted to Multiple.
+        /// </summary>
+        public const string MaxAgents = "MaxAgents";
+
+        /// <summary>
+        /// Indicates how to allocate assemblies to processes. Values are:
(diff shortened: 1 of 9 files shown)


src/nunit.gui.core/Model/TestModel.cs
```diff
@@ -96,32 +96,30 @@ public bool HasResults
...
         {
             get
             {
                 if (_runtimes == null)
                 {
-                    _runtimes = new List<RuntimeFramework>();
-                    foreach (var runtime in RuntimeFramework.AvailableFrameworks)
+                    _runtimes = new List<IRuntimeFramework>();
+                    var availableFrameworks = GetService<IAvailableRuntimes>();
+                    foreach (var runtime in GetService<IAvailableRuntimes>().AvailableRuntimes)
```
immeraufdemhund: looks like you pulled out a variable and forgot to use it here

src/nunit.gui.core/Model/ITestModel.cs
```diff
@@ -78,9 +78,11 @@ public interface ITestModel : IServiceLocator, System.IDisposable
         // See if a test is running
         bool IsTestRunning { get; }
 
+        // Do we have results from running the test?
         bool HasResults { get; }
 
-        IList<RuntimeFramework> AvailableRuntimes { get; }
+        // List of available runtimes
+        IList<IRuntimeFramework> AvailableRuntimes { get; }
```
immeraufdemhund: Comment seems unnecessary

src/nunit.gui.core/Model/TestModel.cs
```diff
@@ -96,32 +96,30 @@ public bool HasResults
...
                     {
                         // Nothing below 2.0
                         if (runtime.ClrVersion.Major < 2)
                             continue;
 
                         // Remove erroneous entries for 4.5 Client profile
-                        if (runtime.FrameworkVersion.Major == 4 && runtime.FrameworkVersion.Minor > 0)
-                            if (runtime.DisplayName.EndsWith("Client"))
-                                continue;
+                        if (runtime.FrameworkVersion.Major == 4 && runtime.FrameworkVersion.Minor > 0 && runtime.Profile == "Client")
```
immeraufdemhund: I feel like for 3 conditions it should be moved to a small method like `if (IsFrameworkErronious(runtime))`. I know the comment exists above it but my eyes glaze over comments usually.

src/nunit.gui.core/Model/TestModel.cs
```diff
@@ -96,32 +96,30 @@ public bool HasResults
...
+        // client and full profiles when executing tests. We
+        // drop unwanted entries here. Even if some of these items
+        // are removed in a later version of the engine, we may
+        // have to retain this code to work with older engines.
+        private List<IRuntimeFramework> _runtimes;
+        public IList<IRuntimeFramework> AvailableRuntimes
         {
             get
             {
                 if (_runtimes == null)
```
immeraufdemhund: Can you switch this to reduce nesting?

src/nunit.gui.core/XmlHelper.cs
```diff
@@ -0,0 +1,165 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Xml;
+
+namespace NUnit
```
immeraufdemhund: I'm assuming this class got copied, do we need to adjust the namespace? or is this on purpose?

src/nunit.gui.core/EnginePackageSettings.cs
```diff
@@ -0,0 +1,157 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit
```
immeraufdemhund: I'm assuming this class got copied. Should the namespace get updated?

CharliePoole: It's sort of related to the model, which is why I put it in this assembly, but it's not really part of it and it's not part of the view or presenter either. Currently only the presenter uses it, but the model might in the future. For all those reasons, I put it in the root NUnit namespace.

src/nunit.gui.core/Model/ITestModel.cs
CharliePoole: Fair enough. Trying to be "consistent". :smile_cat:

src/nunit.gui.core/Model/TestModel.cs
CharliePoole: The pattern is pretty standard for lazy evaluation but I think I'll go back to initializing it in the constructor by calling a method. The list is pretty much always needed anyway.

src/nunit.gui.core/Model/TestModel.cs
CharliePoole: Whoops! Actually, I was using it and changed my mind. Forgot to delete the declaration.

src/nunit.gui.core/Model/TestModel.cs
CharliePoole: OK

src/nunit.gui.core/XmlHelper.cs
CharliePoole: Same as the other one... it's a general class used in all parts of the application. I usually do this until I get up to three or four utility classes and then create a new namespace, but with just a few it's handy to be able to use them without a using statement.

src/nunit.gui.core/Model/TestModel.cs
CharliePoole: Actually, I'll keep the pattern but extract a method.

src/tests/nunit-gui.tests.csproj
```diff
@@ -31,20 +31,12 @@
...
       <HintPath>..\..\packages\NSubstitute.2.0.3\lib\net35\NSubstitute.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="nunit.engine, Version=3.6.1.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.1\lib\nunit.engine.dll</HintPath>
-      <Private>True</Private>
-    </Reference>
     <Reference Include="nunit.engine.api, Version=3.0.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NUnit.Engine.3.6.1\lib\nunit.engine.api.dll</HintPath>
+      <HintPath>..\..\packages\NUnit.Engine.Api.3.6.1\lib\nunit.engine.api.dll</HintPath>
```
immeraufdemhund: I see that the test project itself got updated, but not packages.config.

CharliePoole: That's odd. I used the nuget menu item to do it.

immeraufdemhund:
It looks like it is looking for the nunit-agent that used to be in the package but now is not.

CharliePoole:
Updated packages.config. I don't entirely understand what happened.

I used "Manage NuGet packages for solution..." to update the `nunit-gui` and `nunit.gui.core` projects. When I came back to do it for `nunit.gui.tests`, I used the "Manage NuGet Packages..." menu item on the project. Maybe there's a bug in how this works.

Best answer is probably to double check by looking at the files themselves after using the IDE to manage NuGet packages.

CharliePoole:
OK, I see what's happening. I'll fix and post an explanation for the future somewhere.

CharliePoole:
I dropped the commits that removed the reference from tests. Tests no longer actually need the reference, but we need to have some way to get the engine copied into the output directory. Note that the console doesn't have to do this, because the engine is built as part of the same solution.

There are some quick and dirty ways to get it copied, but I'd rather think about it a bit, so I'm creating a new issue, which I'll assign to myself. For now, the reference in `nunit-gui.tests` should guarantee the presence of the actual engine components.

PR #222, opened 2017-08-27
Issue 96 show notification of assemblies loading

immeraufdemhund:
To address issue #96

### DIFF
--- a/src/nunit.uikit/Controls/LongRunningOperationDisplay.cs
+++ b/src/nunit.uikit/Controls/LongRunningOperationDisplay.cs
@@ -1,0 +1,91 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Drawing;
+using System.Windows.Forms;
+
+namespace NUnit.UiKit.Controls
+{
+    public class LongRunningOperationDisplay : Form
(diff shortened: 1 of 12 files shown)


immeraufdemhund:
Why didn't it auto connect? I had it in my message and in the PR... grrrr

src/nunit.uikit/Controls/LongRunningOperationDisplay.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
```
mikkelbu: There is missing copyright notice here

immeraufdemhund:
>I think that the code looks fine (I'm not a WinForms expert).
However, there is not room enough in the label to present most assemblies. I tried to load a couple of assemblies and since these were located at C:\src\NUnit\nunit\bin\Release\net-4.5 I only got the text "Loading Assemblies: C:\src\NUnit\nunit\bin\Release\net-4.5\nunit.frame" and no more.

>Also it is a bit strange when one loads assemblies with a lot of tests, like nunit.framework.tests.dll from the nunit project, then the label is presented for a couple of seconds. Afterwards the gui is frozen for 5-8 seconds while refreshing a lot. But perhaps we should solve this in another issue?

So question. I did have wordwrap turned on, but the size is fixed meaning that there was some clipping. EG:

image

I could show the file names only which would help in most cases

image

or i could just show the file name of a single file, but for multiple I could just the count

image

immeraufdemhund:
I rebased and went with the 3rd option as it felt the cleanest; however I wont die if I have to change it in favor of what everyone else wants.

immeraufdemhund:
@mikkelbu how does it look for you now?

mikkelbu:
I also like the last one the most, but I'll first have time to look at the actual code/demo it tonight, as current I get a NRE on a clean checkout. The problem is however due to #221 since the code fails in GetAvailableRuntimes.

> System.NullReferenceException was unhandled
>   HResult=-2147467261
>   Message=Object reference not set to an instance of an object.
>   Source=nunit.gui.core
>   StackTrace:
>        at NUnit.Gui.Model.TestModel.GetAvailableRuntimes() in C:\Temp\GUI\nunit-gui\src\nunit.gui.core\Model\TestModel.cs:line 134
>        at NUnit.Gui.Model.TestModel.get_AvailableRuntimes() in C:\Temp\GUI\nunit-gui\src\nunit.gui.core\Model\TestModel.cs:line 106
>        at NUnit.Gui.Presenters.MainPresenter.InitializeMainMenu() in C:\Temp\GUI\nunit-gui\src\nunit-gui\Presenters\MainPresenter.cs:line 144
>        at NUnit.Gui.Presenters.MainPresenter..ctor(IMainView view, ITestModel model) in C:\Temp\GUI\nunit-gui\src\nunit-gui\Presenters\MainPresenter.cs:line 53
>        at NUnit.Gui.Program.Main(String[] args) in C:\Temp\GUI\nunit-gui\src\nunit-gui\Program.cs:line 85
>   InnerException: 


Ps. Is it misleading to write "Assemblies" when we could also be loading an nunit, csproj, sln... file?

CharliePoole:
I like how it looks!

I do see some problems in how it's structured with the rest of the app...

1. The view now knows when tests are loading. There is nothing else like that in the view. For example, the view doesn't know when tests are running, only the presenter knows that and it tells the view what to display.

2. The existing IMessageDisplay in each view wouldn't work for this, obviously. I guess IMainView needs another kind of display for long-running messages. Essentially, you implemented that but gave it a name that implies it can only be used for loading assemblies, so maybe it's just a naming issue.

3. There's a certain logic to putting it in UiKit if we were not trying to get rid of that assembly. 😕 We can remove it as we remove other stuff, however.

immeraufdemhund:
Interesting perception. I see what you are saying. I know in WPF it isn't too hard to put a panel on top of everything that is transparent  and so i'd only show what I want to without the user being able to click or keyboard navigate around. I'll give it some thought this week as I have a pressing matter at work that i need to get out of the door. My first thought is to take the code that I added to the view out. Change the presenter to disable being able to run tests, Then either making a different type of view or message interface make it sit on top of the GUI. I believe we could use this as well just like in V2 for when tests are loading.

PR #225, opened 2017-08-29
fixed failed test with different default culture. Fixes issue #223

immeraufdemhund:


CharliePoole:
@immeraufdemhund We don't want both the issue and the PR in the milestone, just the issue. Only put a PR in a milestone if there is no corresponding issue and it fixes something that should be reported in the release notes.

PR #230, opened 2017-09-09
Create separate assembly for TestModel

### LINKED ISSUE #229: Create separate assembly for the model alone
Progress on issue #189 has given us an `nunit.gui.core` assembly, which mixes the model with the view interfaces. However, the view interfaces are peculiar to the SWF implementation of the gui. We need to create an assembly that contains the common model elements that would be used in any implementation of the GUI, without respect to technical platform or to the shape of the GUI itself.

CharliePoole:
Fixes #229

The new `nunit.testmodel` assembly contains the model only and has no reference to SWF. We may decide to eventually move it to a separate repo, but that would require splitting the tests as well.


src/nunit.testmodel/nunit.testmodel.csproj
```diff
@@ -8,7 +8,7 @@
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Nunit.Gui</RootNamespace>
-    <AssemblyName>nunit.gui.core</AssemblyName>
+    <AssemblyName>nunit.testmodel</AssemblyName>
```
immeraufdemhund: So i've noticed lately that `nunit.testmodel` and `nunit.uikit` are different naming conventions (personally i prefer it this way) compared to `nunit-gui` and `mock-assembly` It throws me off sorta,...and i'm not sure why. Maybe it's the way windows sorts the folders.

ugh.. kids need me. brb for more reviewing.

build.cake
```diff
@@ -136,6 +136,9 @@ Task("PackageZip")
 
         CopyFileToDirectory("LICENSE", BIN_DIR);
         CopyFileToDirectory("CHANGES.txt", BIN_DIR);
+		// Temporary hack... needs update if we update the engine
+		CopyFileToDirectory("packages/NUnit.Engine.3.7.0/lib/nunit-agent.exe.config", BIN_DIR);
+		CopyFileToDirectory("packages/NUnit.Engine.3.7.0/lib/nunit-agent-x86.exe.config", BIN_DIR);
```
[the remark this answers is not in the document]
CharliePoole: Wish I could get VS to do this right!

src/nunit.testmodel/nunit.testmodel.csproj
CharliePoole: Do you mean the hyphens? I use dot-separated components to indicate that the rhs is subordinate to the lhs. So, uikit and testmodel are parts of NUnit. I use hyphens when somebody or thing has a hyphenated name - circular logic, I know. 😄 Like Jean-Jacques or nunit-console. Of course nunit-console and nunit-gui can be construed either way - as a single name or as in "the console component of nunit". But for mock-assembly, there's no doubt that it's a single name, that assembly is not a component of mock, that is. Anyway, it makes sense to me. 😄

CharliePoole: Also, I've never liked dot-separated names for an exe.

PR #234, opened 2017-09-15
Set OutputPath of demo-assembly

mikkelbu:
While building I noticed that the dll of the demo-assembly ended up in another location than all the other dlls.

> ...
> Microsoft (R) Build Engine version 15.1.1012.6693
> Copyright (C) Microsoft Corporation. All rights reserved.
> 
>   nunit.testmodel   -> C:\src\NUnit\nunit-gui\bin\Debug\nunit.testmodel.dll
>   nunit.uikit       -> C:\src\NUnit\nunit-gui\bin\Debug\nunit.uikit.dll
>   nunit-gui         -> C:\src\NUnit\nunit-gui\bin\Debug\nunit-gui.exe
>   mock-assembly     -> C:\src\NUnit\nunit-gui\bin\Debug\mock-assembly.dll
>   nunit-gui.tests   -> C:\src\NUnit\nunit-gui\bin\Debug\nunit-gui.tests.dll
>   demo-assembly     -> C:\src\NUnit\nunit-gui\src\demo-assembly\bin\Debug\demo-assembly.dll

Hence, the OutputPath of the demo-assembly is now set to the same folders as the other projects.

### DIFF
--- a/src/demo-assembly/demo-assembly.csproj
+++ b/src/demo-assembly/demo-assembly.csproj
@@ -20,7 +20,7 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>..\..\bin\Debug\</OutputPath>
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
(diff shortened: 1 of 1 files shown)


CharliePoole:
I kept it in a different location in order to simulate the experience a user would have. In the past, I've run into bugs that users saw, but we didn't, because all our assemblies were in the same directory as the test assembly.

mikkelbu:
Ok. I thought that it was an error. I'll close the PR.

PR #240, opened 2017-11-14
Separate C# GUI initialization logic from the rest of the entry point

ErichDonGubler:
This simple PR is intended to allow another program to act as a wrapper for the GUI runner. Currently, one can invoke the runner via inflection, but if one uses C# forms BEFORE invoking the runner, the GUI initialization logic in `Main` throws an exception. In my case, for instance, I was initializing a C# form for easy configuration of the args passed to the NUnit GUI runner, and the following exception would get thrown upon using reflection to call `Main`:

```
Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.InvalidOperationException: SetCompatibleTextRenderingDefault must be called before the first IWin32Window object is created in the application.
   at System.Windows.Forms.Application.SetCompatibleTextRenderingDefault(Boolean defaultValue)
   at NUnit.Gui.Program.Main(String[] args)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
// ...
```

With this change, clients can now invoke the 

```
var guiRunnerAssembly = Assembly.LoadFile("...\\nunit-gui.exe");
guiRunnerAssembly.EntryPoint.DeclaringType.GetMethod("Run").Invoke(null, new object[] { args });
```

### DIFF
--- a/src/nunit-gui/Program.cs
+++ b/src/nunit-gui/Program.cs
@@ -43,7 +43,11 @@ namespace NUnit.Gui
         {
             Application.EnableVisualStyles();
             Application.SetCompatibleTextRenderingDefault(false);
+            Run(args);
+        }
 
+        public static void Run(string[] args)
+        {
             CommandLineOptions options = new CommandLineOptions();
             try
             {


CharliePoole:
I don't see any particular harm in this PR, but it supports a usage that we don't have any plans to continue to support. That is, it could continue to work, or not, depending on future changes.

@NUnitSoftware/gui-team What do you think?

ErichDonGubler:
Well, if it's not something that the team *plans* on supporting, I definitely don't want to give a false start that makes it *seem* like it's supported. :) I hope that the GUI team likes it!

For me as a user, the motivation behind this example is to avoid shelling out using `Process.Start`, if possible. It's nice to avoid that overhead, but that doesn't change the fact that it adds another concern to the GUI runner's code: maintaining the entry point as I've modified it here.

CharliePoole:
For the V2 gui, we made the exe a one-liner and put the actual code in an eye that others could reference. Of course, you can reference the exe as well.

I wonder if a more robust fix would be possible by bypassing initialization if it's already done by a caller.

mikkelbu:
I don't mind the change. Wrt. to making the code more robust I'm unsure about the consequences of `SetCompatibleTextRenderingDefault`. What if the caller has set `Application.SetCompatibleTextRenderingDefault(true)` before calling into our code. How will this affect our code/GUI?

immeraufdemhund:
That part wasn't changed, and it would be up to the user who is bypassing to deal with the consequences of setting `Application.SetCompatibleTextRenderingDefault(true)`.  I think the change is small enough and fine as is.

src/nunit-gui/Program.cs
```diff
@@ -43,7 +43,11 @@ static void Main(string[] args)
         {
             Application.EnableVisualStyles();
             Application.SetCompatibleTextRenderingDefault(false);
+            Run(args);
+        }
 
+        public static void Run(string[] args)
```
CharliePoole: Would you add a comment about why we have this separate entry?

ErichDonGubler: More than happy to. :)

ErichDonGubler: Done. How's that?

ErichDonGubler:
Also, another thing I might add for the benefit of anyone who's following this: being able to call the executable via assembly loading instead of shelling out also facilitates a debugging flow from Visual Studio. I just ran into this myself today -- I'd forgotten that this was one of the ramifications!

PR #242, opened 2017-11-28
Only set TestXml when relevant

mikkelbu:
TestXml is now set when:
* We change from the tab "Properties" to the tab "XML"
* Or when the tab "XML" is visible and the selected item in the tree changes

This improves the performance of most operations in the gui
(i.e. all operations as long as the "XML" tab is not selected). The only
downside is that going back and forth between the tabs will retrigger
a call to `DisplayXml`.

To improve the experience even more `GetFullXml` and `XmlHelper.ToRtfString`
could be done asynchronous and results could be cached.

Relates to #241

### DIFF
--- a/src/nunit-gui/Views/MainForm.cs
+++ b/src/nunit-gui/Views/MainForm.cs
@@ -189,5 +189,11 @@ namespace NUnit.Gui.Views
             if (files != null)
                 DragDropFiles?.Invoke(files);
         }
+
+        private void TabControlSelectedIndexChanged(object sender, System.EventArgs e)
+        {
+            if (tabControl1.SelectedIndex == 1)
+                XmlView.InvokeFocus();
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


CharliePoole:
While it could still be improved, I think this makes sense to merge right now. In the long term, I like your idea of making it a right-click dialog, maybe with some caching, and finally removing it entirely once we are done developing.

Even though you can use the debugger, I for one find it convenient to have this when working on the UI... if what I see displayed doesn't make sense, then I can quickly look at the XML from which it is drawn.

src/nunit-gui/Views/MainForm.cs
```diff
@@ -189,5 +189,11 @@ protected override void OnDragDrop(DragEventArgs drgevent)
             if (files != null)
                 DragDropFiles?.Invoke(files);
         }
+
+        private void TabControlSelectedIndexChanged(object sender, System.EventArgs e)
+        {
+            if (tabControl1.SelectedIndex == 1)
```
immeraufdemhund: I know you said not to merge, but can this be changed to check that the selected tab is the xml tab and not based off of the index?

mikkelbu: Done. Please review 😄

CharliePoole: I'm all for it. Merge when ready!

PR #243, opened 2017-11-29
Avoid "The path is not of a legal form" GUI crash

mikkelbu:
Note that we currently do not show top-level failures
(a failure-node just inside the outermost test-suite-node)
in the GUI (except in the XML tab), but at least the GUI
does not crash anymore.

Relates to #228 and #184


CharliePoole:
Thanks!

PR #262, opened 2018-11-17
Run tests under Travis CI

CharliePoole:
Up to now, we didn't run tests in Travis at all.There were only a few fixes having to do with Newline plus a large number of tests of the Views that create the view classes without properly initializing them. This works for our tests in Windows but not Linux.

### DIFF
--- a/src/TestCentric/tests/Views/CommonViewTests.cs
+++ b/src/TestCentric/tests/Views/CommonViewTests.cs
@@ -30,6 +30,7 @@ namespace TestCentric.Gui.Views
 {
     [TestFixture(typeof(MainForm))]
     [TestFixture(typeof(TestTreeView))]
+	[Platform(Exclude = "Linux", Reason = "Uninitialized form causes an error in Travis-CI")]
     public class CommonViewTests<T> where T: new()
     {
         protected T View { get; private set; }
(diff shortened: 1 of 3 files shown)


src/TestCentric/tests/Views/CommonViewTests.cs
```diff
@@ -30,6 +30,7 @@ namespace TestCentric.Gui.Views
 {
     [TestFixture(typeof(MainForm))]
     [TestFixture(typeof(TestTreeView))]
+	[Platform(Exclude = "Linux", Reason = "Uninitialized form causes an error in Travis-CI")]
```
mikkelbu: @CharliePoole I just noticed that this line (and the change below) is inserted using a tab and not spaces. Would it make sense also to clean up all the files in the `testcentric-experimental-gui` and if so should be do it before or after combining the projects.

CharliePoole: I think that's a good idea. Let's add a note to the project to that effect. (I'm using notes rather than issues unless it's something that needs to show up in an eventual readme.)

This example gives me a hint of how it's happening. IIRC, I typed in the first Platform attribute, then highlighted and shift-deleted the line, then pasted it back and pasted it back a second time in the other file. That's a pattern I have. So, somewhere in that process it's not being formatted correctly. I'll check the project settings.

If you want to do the reformatting, there should be a point of quiescence shortly.

PR #297, opened 2019-07-26
Update README.md

mikkelbu:


mikkelbu:
I'm merging even though Travis is failing. I looks like the configuration needs to be updated.

